PR #2, opened 2016-03-26
Fix ng-admin with api-platformv2

Simperfit:
Make the ng-admin work by test if the the supported property exists


dunglas:
Thanks @Simperfit

Issue #5, opened 2017-03-31
Publish hydra client as a standalone package


fzaninotto:
Hi guys, great package!

Do you think it makes sense to publish the `hydraClient` as a standalone npm package, and mention it in https://marmelab.com/admin-on-rest/RestClients.html?

dunglas:
Hi @fzaninotto,

There just a few files in this repository and the Hydra client has been designed to be usable without other features. Components provided by this package can also be used one by one with an existing admin on rest installation.
Is it worth it to split this package in 2?

Regarding the doc, I'll send a PR!

fzaninotto:
It's up to you!

trsteel88:
Are there any example docs to use the hydra client standalone? I want more control over all the layouts and would prefer to use react-admin + the hydraClient. I don't like all the magic that the api platform admin performs.

dunglas:
This is now documented @trsteel88: https://api-platform.com/docs/admin/getting-started#using-the-hydra-data-provider-directly-with-react-admin

trsteel88:
Thanks. I saw this about an hour ago and thought I was losing my mind and missed it today.

trsteel88:
Within my app I have a password reset endpoint. Is there any documentation for how I would load that in as a custom route?

Sorry for all the questions. This is a great library and you've done a great job. I'd be happy to contribute to the docs once I have a better understanding.

trsteel88:
Hey @dunglas, are you able to help me with an issue I'm having with a ReferenceArrayInput?

When creating the form manually, I get the following error:

```
Uncaught TypeError: Cannot read property 'data' of undefined
    at possibleValues.js:68
    at Array.map (<anonymous>)
    at getPossibleReferences (possibleValues.js:67)
    at ReferenceArrayInputController.js:349
    at index.js:76
    at index.js:37
    at index.js:89
    at Function.mapToProps (index.js:37)
    at mapToPropsProxy (wrapMapToProps.js:41)
    at handleNewState (selectorFactory.js:55)
    at handleSubsequentCalls (selectorFactory.js:69)
    at pureFinalPropsSelector (selectorFactory.js:74)
    at Object.runComponentSelector [as run] (connectAdvanced.js:72)
    at Connect.onStateChange (connectAdvanced.js:239)
    at Object.notify (Subscription.js:29)
    at Subscription.notifyNestedSubs (Subscription.js:66)
    at Connect.onStateChange (connectAdvanced.js:242)
[...]

PR #7, opened 2017-04-07
[WIP] Feature/prop types refactoring

lutangar:
Hi there,

I'm adding better and more flexible `propTypes` validation based on existing  [admin-on-rest](https://github.com/marmelab/admin-on-rest/blob/master/src/Admin.js#L116) `propTypes`.
This way you won't have to maintain propTypes that you don't own.

It also fix customization issues of an `Admin` inside an `AdminBuilder` by passing down all props to child component except the custom ones.

I could go further down this road, for example by creating a dedicated `customPropTypes.js` file and reduce duplication:
```
export const options = {
  inputFactory: React.PropTypes.func,
  resource: React.PropTypes.instanceOf(Resource),
})
```
That's currently a WIP, but I'd like to hear from you before moving any further.
Also what's your plan for this project? Do you plan to split this up as suggested in #5 ?

Cheers.

### DIFF
--- a/src/AdminBuilder.js
+++ b/src/AdminBuilder.js
@@ -10,23 +10,25 @@ import Edit from './Edit';
 export default class AdminBuilder extends Component {
   static propTypes = {
     api: React.PropTypes.instanceOf(Api).isRequired,
-    restClient: React.PropTypes.func.isRequired,
+    ...Admin.propTypes,
   };
 
   render() {
-    return <Admin title={this.props.api.title} restClient={this.props.restClient}>
-        {this.props.api.resources.map(resource =>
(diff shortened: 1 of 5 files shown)


dunglas:
Hi @lutangar, sorry for the delay. I've just replied to #5. There is no plan to split the package yet.
Do you think you can finish this PR?

dunglas:
@lutangar any news regarding this PR?

lutangar:
Hey @dunglas, I'm not into this at the moment, feel free to take it or close it if you want to.

dunglas:
Closed by #63.

PR #12, opened 2017-05-19
Allow transformation of embedded json-ld document

lutangar:
Hey,

The `transformJsonLdToAOR` has been updated to handle embedded document recursively.
This function has also been exposed to be easily tested.


dunglas:
Thank you @lutangar!

PR #15, opened 2017-06-16
Upgrade to Admin On Rest 1.1

dunglas:
ping @mauchede


src/inputFactory.js
```diff
@@ -30,7 +32,7 @@ export default (input) => {
       break;
 
     case 'http://www.w3.org/2001/XMLSchema#boolean':
-      delete validation.required;
+      if (!input.validate && input.required) validate = undefined;
```
dunglas: I'll improve this in another PR.

PR #24, opened 2017-06-26
Allow to inject props in fields and inputs

dunglas:


README.md
```diff
@@ -191,6 +191,44 @@ export default Admin;
 Refer to [the chapter dedicated to authentication in the Admin On Rest documentation](https://marmelab.com/admin-on-rest/Authentication.html)
 for more information.
 
+## Use a Custom Validation Function or Inject Custom Props
+
+You can use `fieldProps` and `inputProps` to respectively inject custom properties to fields and inputs generated by API
+Platform Admin. This is particularly useful to add custom validation rules:
```
Simperfit: This become useful when you have to add custom validation rules:

src/fieldFactory.js
```diff
@@ -34,17 +34,15 @@ export default (field) => {
   }
 
   if (null !== field.reference) {
-    return <ReferenceField source={field.name}
-                           reference={field.reference.name} key={field.name}>
+    return <ReferenceField source={field.name} reference={field.reference.name} key={field.name} {...field.fieldProps}>
       <TextField source="id"/>
-    </ReferenceField>
+    </ReferenceField>;
```
Simperfit: I don't think we need this

PR #27, opened 2017-06-29
Fix JsonLd to AOR document transformation 

lutangar:
Embedded resources and collections are now transformed correctly.

Previous tests were flawed because `transformJsonLdToAOR` was mutating the original document.
Implementation has been changed to deal with this and tests were added to ensure that each resource of a collection are also transformed.


dunglas:
Thank you very much for fixing this @lutangar

PR #33, opened 2017-07-24
Fix deletion

mauchede:
`admin-on-rest` requires a `data` property in response but `DELETE` response does not contain any `data` node.

In this fix, I add a default value to `convertAORDataToHydraData` and `convertHydraDataToAORData`. If `data` node is missing (it will be `undefined`), `{}` will be used as value of `data`.


schornio:
+1

dunglas:
Thanks @mauchede

PR #36, opened 2017-08-07
add build

ginifizz:


meyerbaptiste:
Why?

dunglas:
The build should not be versioned (it makes searching the repo harder). It is included in the NPM package. 👎 for me.

ginifizz:
Wooops...

PR #45, opened 2017-08-29
Fix HydraAdmin authentifaction documentation

jbeurel:
It seems that the entrypoint prop is required

https://github.com/api-platform/admin/blob/master/src/hydra/HydraAdmin.js#L15


Simperfit:
Thanks @jbeurel

Issue #43, opened 2017-08-26
Admin Does Not Require authentication for docs.jsonld request


playmice:
I noticed that when the whole symfony project is under a security layer, the Admin won't load because docs.jsonld returns 403 error. Normally, the 403 error should take you to the login page, but it doesn't in the case of docs.jsonld.
The only workaround for now is to open up docs.jsonld to anonymous users, which isn't good from a security perspective.

dunglas:
Closed by #51.

Issue #31, opened 2017-07-12
HydraAdmin component isn't working with no entrypoint props


Lctrs:
Hi,

I followed the docs to implement the admin with JWT Authentication support. Here is the code :

```js
// authClient.js
import { AUTH_LOGIN, AUTH_LOGOUT, AUTH_ERROR, AUTH_CHECK } from 'admin-on-rest';

export default (type, params) => {
  if (type === AUTH_LOGIN) {
    const { username, password } = params;
    const request = new Request(`${process.env.REACT_APP_ENTRYPOINT}/login_check`, {
      method: 'POST',
      body: JSON.stringify({ username, password }),
      headers: new Headers({ 'Content-Type': 'application/json' }),
    });

    return fetch(request)
      .then(response => {
        if (response.status < 200 || response.status >= 300) {
          throw new Error(response.statusText);
        }

        return response.json();
      })
      .then(({ token }) => {
        localStorage.setItem('token', token);
      });
  }

  if (type === AUTH_LOGOUT) {
    localStorage.removeItem('token');

    return Promise.resolve();
  }

  if (type === AUTH_ERROR) {
    const { status } = params;
    if (status === 401 || status === 403) {
      localStorage.removeItem('token');

      return Promise.reject();
    }

    return Promise.resolve();
  }

  if (type === AUTH_CHECK) {
    return localStorage.getItem('token') ? Promise.resolve() : Promise.reject();
  }

  return Promise.reject('Unknown method.');
};
```

```js
// App.js
import React, { Component } from 'react';
import { HydraAdmin, hydraClient, fetchHydra } from 'api-platform-admin';
import authClient from './authClient';

const fetchWithAuth = (url, options = {}) => {
  if (!options.headers) {
    options.headers = new Headers({ Accept: 'application/ld+json' });
  }

  options.headers.set(
    'Authorization',
    `Bearer ${localStorage.getItem('token')}`
  );
  return fetchHydra(url, options);
};

class App extends Component {
  render() {
    return (
      <HydraAdmin
        restClient={hydraClient(
          process.env.REACT_APP_ENTRYPOINT,
          fetchWithAuth
        )}
        authClient={authClient}
      />
    );
  }
}

export default App;
```

Problem is that `HydraAdmin` component requires `entrypoint` prop to be set as it passes it to `parseHydraDocumentation()` in `componentDidMount()`. But the doc says that it is not required in this case.

So, I added it to my App's render method :

```js
// App.js
// ...
class App extends Component {
[...]

PR #51, opened 2017-09-06
Handle error during api documentation parsing

mauchede:
This PR should fix issues #43 and #31.

---

This PR handles parsing errors (see [my PR on dunglas/api-doc-parser](https://github.com/dunglas/api-doc-parser/pull/13)):
* if the parsing is resolved, the "classic" behavior is kept (rendering `AdminBuilder`).
* if the parsing is rejected, a message will be shown.

In all cases, the user has the possibility to define some `customRoutes`. These routes will be added to the `customRoutes` defined via properties.

---

With this PR, we can now use `api-platform/admin` on a fully secured API. For example:

```js
import { AUTH_ERROR, AUTH_LOGIN, AUTH_LOGOUT } from 'admin-on-rest';
import parseHydraDocumentation from 'api-doc-parser/lib/hydra/parseHydraDocumentation';
import { fetchHydra as baseFetchHydra, HydraAdmin, hydraClient as baseHydraClient } from 'api-platform-admin';
import React from 'react';
import { Redirect } from 'react-router-dom';

// Define constants

const ENTRYPOINT = 'http://127.0.0.1:8000';
const LOCAL_STORAGE_KEY_TOKEN = 'token';

// Define restClient

const fetchHeaders = {
    'Authorization': `Bearer ${window.localStorage.getItem(LOCAL_STORAGE_KEY_TOKEN)}`,
};

const fetchHydra = (url, options = {}) => baseFetchHydra(url, {
    ...options,
    headers: new Headers(fetchHeaders),
});

const restClient = api => baseHydraClient(api, fetchHydra);

// Define authClient

const authClient = (type, params) => {
    switch (type) {
        case AUTH_ERROR:
            if (401 === params.status || 403 === params.status) {
                window.localStorage.removeItem(LOCAL_STORAGE_KEY_TOKEN);
                window.location.reload();
                break;
            }

            return Promise.resolve();

        case AUTH_LOGIN:
            const body = new FormData();
            body.append('username', params.username);
            body.append('password', params.password);

            const request = new Request(`${ENTRYPOINT}/login`, {
                method: 'POST',
                body,
            });

            return fetch(request)
                .then(response => {
                    if (response.status < 200 || response.status >= 300) {
                        throw new Error(response.statusText);
                    }

                    return response.json();
                })
                .then(({ token }) => {
                    window.localStorage.setItem(LOCAL_STORAGE_KEY_TOKEN, token);
                    window.location.replace('/');
                });

        case AUTH_LOGOUT:
            window.localStorage.removeItem(LOCAL_STORAGE_KEY_TOKEN);

            return Promise.resolve();

        default:
            return Promise.resolve();
    }
};

// Define apiDocumentationParser

const apiDocumentationParser = entrypoint => parseHydraDocumentation(entrypoint, { headers: new Headers(fetchHeaders) })
    .then(
        ({ api }) => ({ api }),
        (result) => {
            switch (result.status) {
                case 401:
                    return Promise.resolve({
                        api: result.api,
                        customRoutes: [
                            {
                                props: {
                                    path: '/',
                                    render: () => (
                                        <Redirect to={`/login`}/>
                                    ),
                                },
                            },
                        ],
                    });

                default:
                    return Promise.reject(result);
            }
        },
    );


// Render administration

export default props => (
    <HydraAdmin
        apiDocumentationParser={apiDocumentationParser}
        authClient={authClient}
        entrypoint={ENTRYPOINT}
        restClient={restClient}
    />
);
```

In this example:
* we add a `customRoute` if parsing returned a `401` error. This route will redirect to the `login` page.
* the `authClient` will refresh current page when a JWT token will be stored. Refresh the page will give the possibility to parse again the documentation, but this time the JWT token will be sent.

### DIFF
--- a/src/hydra/HydraAdmin.js
+++ b/src/hydra/HydraAdmin.js
@@ -7,34 +7,70 @@ import restClient from './hydraClient';
 class HydraAdmin extends Component {
   static defaultProps = {
     apiDocumentationParser,
+    customRoutes: [],
+    error: 'Unable to retrieve API documentation.',
+    loading: 'Loading...',
     restClient,
   };
 
   static propTypes = {
     apiDocumentationParser: PropTypes.func,
+    customRoutes: PropTypes.array,
     entrypoint: PropTypes.string.isRequired,
+    error: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+    loading: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
     restClient: PropTypes.func,
   };
 
   state = {
     api: null,
+    customRoutes: [],
+    hasError: false,
+    loaded: false,
   };
 
   componentDidMount() {
     this.props
       .apiDocumentationParser(this.props.entrypoint)
-      .then(api => this.setState({api}));
+      .then(
+        ({api, customRoutes = []}) => ({
+          api,
+          customRoutes,
+          hasError: false,
+          loaded: true,
+        }),
+        ({api, customRoutes = []}) => ({
+          api,
+          customRoutes,
+          hasError: true,
+          loaded: true,
+        }),
+      )
+      .then(this.setState.bind(this));
   }
 
   render() {
-    if (null === this.state.api) {
-      return <span>Loading...</span>;
+    if (false === this.state.loaded) {
+      return 'function' === typeof this.props.loading
+        ? <this.props.loading />
+        : <span className="loading">
+            {this.props.loading}
+          </span>;
+    }
+
+    if (true === this.state.hasError) {
+      return 'function' === typeof this.props.error
+        ? <this.props.error />
+        : <span className="error">
+            {this.props.error}
+          </span>;
     }
 
     return (
       <AdminBuilder
         {...this.props}
         api={this.state.api}
+        customRoutes={this.props.customRoutes.concat(this.state.customRoutes)}
         restClient={this.props.restClient(this.state.api)}
       />
     );
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "admin-on-rest": "^1.1",
-    "api-doc-parser": "^0.1.7",
+    "api-doc-parser": "^0.2.0",
     "babel-runtime": "^6.23",
     "jsonld": "^0.4",
     "lodash.isplainobject": "^4.0.6",
--- a/yarn.lock
+++ b/yarn.lock
@@ -115,9 +115,9 @@ anymatch@^1.3.0:
     arrify "^1.0.0"
     micromatch "^2.1.5"
 
-api-doc-parser@^0.1.7:
-  version "0.1.7"
-  resolved "https://registry.yarnpkg.com/api-doc-parser/-/api-doc-parser-0.1.7.tgz#b55697a92b425b2118e7dcd14f9d444b7e10f673"
+api-doc-parser@^0.2.0:
+  version "0.2.0"
+  resolved "https://registry.yarnpkg.com/api-doc-parser/-/api-doc-parser-0.2.0.tgz#7751ec1b7b377cd91be7698b360b7829bb234c33"
   dependencies:
     babel-runtime "^6.23.0"
     jsonld "^0.4.11"


src/hydra/HydraAdmin.js
```diff
@@ -7,34 +7,57 @@ import restClient from './hydraClient';
...
+          hasError: true,
+          loaded: true,
+        }),
+    );
   }
 
   render() {
-    if (null === this.state.api) {
+    if (false === this.state.loaded) {
       return <span>Loading...</span>;
```
Gregcop1: In order to make this more configurable and user friendly, it could be possible to pass messages (loading, error, ...) on props with default values. So my french interface can keep french messaging

Gregcop1: you should add a className here to allow customization of the render

src/hydra/HydraAdmin.js
```diff
@@ -7,34 +7,57 @@ import restClient from './hydraClient';
...
     }
 
+    if (true === this.state.hasError) {
+      return <span>Impossible to retrieve API documentation.</span>;
```
Gregcop1: you should add a className here to allow customization of the render

src/hydra/HydraAdmin.js
```diff
@@ -7,34 +7,57 @@ import restClient from './hydraClient';
...
+    loaded: false,
   };
 
   componentDidMount() {
-    this.props
-      .apiDocumentationParser(this.props.entrypoint)
-      .then(api => this.setState({api}));
+    this.props.apiDocumentationParser(this.props.entrypoint).then(
+      ({api, customRoutes = []}) =>
+        this.setState({
```
Simperfit: this feels duplicated, is it possible to not duplicate it or it's not worth the complexity ?

src/hydra/HydraAdmin.js
Simperfit: maybe this could be a key instead of directly english ?

src/hydra/HydraAdmin.js
```diff
@@ -7,34 +7,57 @@ import restClient from './hydraClient';
...
+    customRoutes: [],
+    hasError: false,
+    loaded: false,
   };
 
   componentDidMount() {
-    this.props
-      .apiDocumentationParser(this.props.entrypoint)
-      .then(api => this.setState({api}));
+    this.props.apiDocumentationParser(this.props.entrypoint).then(
```
Gregcop1: why `.then` on the same line?

src/hydra/HydraAdmin.js
mauchede: Instead of call `setState` directly, we could "normalize" the data:

```js
    this.props
      .apiDocumentationParser(this.props.entrypoint)
      .then(
        ({api, customRoutes = []}) => ({
          api,
          customRoutes,
          hasError: false,
          loaded: true,
        }),
        ({api, customRoutes = []}) => ({
          api,
          customRoutes,
          hasError: true,
          loaded: true,
        }),
      )
      .then(state => this.setState(state));
```

src/hydra/HydraAdmin.js
dunglas: Allowing to pass props as suggested by @Gregcop1 looks better to me.

src/hydra/HydraAdmin.js
```diff
@@ -7,34 +7,70 @@ import restClient from './hydraClient';
...
+          loaded: true,
+        }),
+        ({api, customRoutes = []}) => ({
+          api,
+          customRoutes,
+          hasError: true,
+          loaded: true,
+        }),
+      )
+      .then(state => this.setState(state));
```
Gregcop1: Have you tried `.then(this.setState)` ?

mauchede: I tried and the following error occurred:

```
Uncaught (in promise) TypeError: Cannot read property 'updater' of undefined
```

mauchede: It works, if I use:

```js
.then(this.setState.bind(this))
```

dunglas:
Thanks @mauchede!

Superprogrameur59:
It could be very useful to put this example in the official doc of api platform because this "https://api-platform.com/docs/admin/authentication-support" doc is wrong ;)

PR #53, opened 2017-09-14
Add support for react >= 15.6

Lctrs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
     "eslint-plugin-prettier": "^2.1.2",
     "eslint-plugin-react": "^7.1",
     "jest": "^20.0",
-    "prettier": "^1.5.2"
+    "prettier": "^1.5.2",
+    "raf": "^3.3.2"
   },
   "dependencies": {
     "admin-on-rest": "^1.3",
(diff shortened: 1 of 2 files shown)


Lctrs:
Should I also bump `admin-on-rest` dep constraint to `^1.3` to avoid potential same problem with previous versions ?

soyuka:
> Should I also bump admin-on-rest dep constraint to ^1.3 to avoid potential same problem with previous versions ?

You can.

Lctrs:
Do you have any idea why it fails now ? `react` is listed in the lockfile so I don't see why it can't find it.

soyuka:
Okay so my bad, please keep the react hard dependencies and remove peer dependencies. Indeed `admin-on-rest` has them under `devDependencies` and `peerDependencies`. This way, it won't be installed twice in production builds.

dunglas:
React as a peer dependency is usually a bad idea: https://github.com/yahoo/react-intl/issues/79

Lctrs:
@soyuka Is that what you had in mine ?

soyuka:
Yes. Keep the same constraint on react versions as admin-on-rest though `^15.4.0 || ^16.0.0-beta.5`

soyuka:
Thanks @Lctrs

PR #66, opened 2017-11-02
Add support for UrlField

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo


src/fieldFactory.js
```diff
@@ -26,8 +27,15 @@ export default (field, options) => {
...
-  if ('http://schema.org/email' === field.id) {
-    return <EmailField key={field.name} source={field.name} />;
+  switch (field.id) {
+    case 'http://schema.org/email':
+      return <EmailField key={field.name} source={field.name} />;
+
+    case 'http://schema.org/url':
+      return <UrlField key={field.name} source={field.name} />;
+
+    default:
```
soyuka: you don't really need the default then.

dunglas: If I don't add it ESLint emits a warning.

PR #69, opened 2017-11-09
Remove react from dependency and add it to devDependencies

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

React should not be in `dependencies` but in `devDependencies`.
This should resolve the issue "You may have multiple copies of React loaded." when updating to React 16.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,9 @@
     "eslint-plugin-react": "^7.1",
     "jest": "^20.0",
     "prettier": "^1.5.2",
-    "raf": "^3.3.2"
+    "raf": "^3.3.2",
+    "react": "^15.6.1",
+    "react-dom": "^15.6.1"
   },
   "dependencies": {
     "@api-platform/api-doc-parser": "^0.3.1",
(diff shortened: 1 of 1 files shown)


dunglas:
The  problem is that AOR and MUI aren’t compatible with React 16 yet and use the dependency strategy.

alanpoulain:
Are you sure for AOR and MUI? Because their package.json files seem OK to me:
https://github.com/marmelab/admin-on-rest/blob/07452f0efe31fda3309710f74e11f5988595206c/package.json
https://github.com/callemall/material-ui/blob/f1bc4be26ca68578f0cfd053fed8eeabe0d3a636/package.json
And when I have done this modification, I succeeded in loading React 16 in the project.

dunglas:
Not in the stable version: https://github.com/marmelab/admin-on-rest/blob/v1.3.2/package.json#L48

alanpoulain:
This is a devDependencies so it's good. It's only an issue when it's in the dependencies.

dunglas:
Thanks @alanpoulain

PR #80, opened 2018-02-22
Extract logic from AdminBuilder to export resourceFactory

k911:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

It would be easier create custom Admin app (not using `admin-on-rest`'s Admin component), or inject into existing one.

Todo:
- [x] export function
- [x] add test

### DIFF
--- a/src/AdminBuilder.js
+++ b/src/AdminBuilder.js
@@ -1,48 +1,19 @@
 import Api from '@api-platform/api-doc-parser/lib/Api';
-import {Admin, Delete, Resource} from 'admin-on-rest';
+import {Admin} from 'admin-on-rest';
 import PropTypes from 'prop-types';
 import React from 'react';
-import Create from './Create';
-import Edit from './Edit';
 import fieldFactory from './fieldFactory';
 import inputFactory from './inputFactory';
-import List from './List';
-import Show from './Show';
+import resourceFactory from './resourceFactory';
 
 const AdminBuilder = props => {
   const {api, fieldFactory, inputFactory, title = api.title} = props;
 
   return (
     <Admin {...props} title={title}>
-      {api.resources.map(resource => {
-        const {
-          create = Create,
-          edit = Edit,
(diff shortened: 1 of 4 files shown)


dunglas:
Looks good to me, can you just add a test for this new function please?

k911:
@dunglas Yeah sure, I'll update this as soon as possible

src/resourcesBuilder.js
```diff
@@ -0,0 +1,51 @@
+import {Delete, Resource} from 'admin-on-rest';
```
mauchede: We already have two factories: `inputFactory` and `fieldFactory`. I suggest you to rename this file `resourceFactory`.

src/resourcesBuilder.js
```diff
@@ -0,0 +1,51 @@
...
+      show={show}
+    />
+  );
+};
+
+/**
+ * Creates array of Resource components from Api instance 
+ * using provided input/field factories or default ones
+ */
+export default (
```
mauchede: This function should be in `AdminBuilder`:
* `AdminBuilder` already have the default value of `fieldFactory` and `inputFactory`.
* behaviour will be the same for all factories.

src/resourcesBuilder.js
k911: I'll refer to your both comments here, IMO factory should return only one item and obviously there it would return array of `Resource` components. Maybe then we should rename it indeed to `resourceFactory` but export only what's currently in code `const makeResource`? WDYT?

k911:
@mauchede I've updated PR, hope that's what you expected. 

I've also created some tests, please verify them because I'm newbie to js/react.

dunglas:
Thanks @k911

PR #92, opened 2018-04-17
Add ability to customize request body type

k911:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

**Why?**

I've tried to create decorator of `fetchHydra` to handle additional API Endpoint which uses `multipart/form-data` content type as request body to send a file, but I've encountered a problem, which made me to copy whole `httpClient` from this package and add some logic. It is all because in `fetchApi` method of `httpClient` we first `convertAORRequestToHydraRequest` where simply `JSON.stringify()` is done, then use decorated `httpClient`.

I figured out, that if we're able to normalize all property fields, we could also serialize whole resource. It may be helpful not only for passing `FormData` object as request body but also gives ability to use custom object serializer.

Also please note that in `@api-platform/api-doc-parser` we already support sending body as `FormData`.

I'll create doc PR if you agree to this feature. 

In meantime this is a example usage while decorating `parseHydraDocumentation`:

```javascript
// assert 'image' field of 'resource' is ImageInput type and normalizeData on this field returns `File` object

resource.serializeData = (data) => {
  const body = new FormData();
  body.append('image', data.image);
  return body;
};
```
Then automatically when calling CREATE / UPDATE action resource is handled properly, even without additional configuration.

### DIFF
--- a/src/hydra/hydraClient.js
+++ b/src/hydra/hydraClient.js
@@ -65,17 +65,12 @@ export const transformJsonLdDocumentToAORDocument = (
  */
 export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
   /**
-   * @param {string} resource
+   * @param {Object} resource
    * @param {Object} data
    *
    * @returns {Promise}
    */
   const convertAORDataToHydraData = (resource, data = {}) => {
-    resource = resources.find(({name}) => resource === name);
-    if (undefined === resource) {
-      return Promise.resolve(data);
-    }
-
     const fieldData = [];
     resource.fields.forEach(({name, normalizeData}) => {
       if (!(name in data) || undefined === normalizeData) {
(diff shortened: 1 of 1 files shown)


src/hydra/hydraClient.js
```diff
@@ -98,6 +93,25 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
...
+   * @returns {Promise}
+   */
+  const transformAORDataToRequestBody = (resource, data = {}) => {
+    resource = resources.find(({name}) => resource === name);
+    if (undefined === resource) {
+      return Promise.resolve(data);
+    }
+
+    return convertAORDataToHydraData(resource, data).then(data => {
+      return undefined === resource.serializeData
```
mauchede: Could you rename `serializeData` to `encodeData`? The logic is pretty same than [Serializer component](https://symfony.com/doc/current/components/serializer.html).

src/hydra/hydraClient.js
```diff
@@ -65,17 +65,12 @@ export const transformJsonLdDocumentToAORDocument = (
...
    *
    * @returns {Promise}
    */
-  const convertAORDataToHydraData = (resource, data = {}) => {
-    resource = resources.find(({name}) => resource === name);
-    if (undefined === resource) {
-      return Promise.resolve(data);
-    }
-
+  const convertAORDataToHydraData = (resource, data) => {
```
mauchede: A default value is needed for data (see #33).

src/hydra/hydraClient.js
k911: :+1: That's very good name!

k911:
Thanks @mauchede, I've updated PR.
Yeah, that's true. It's basically my main purpose of this PR: simplify handling of uploading images :)

mauchede:
It is OK for me. Thank you @k911!

PR #95, opened 2018-05-28
Exclude deprecated resources and fields

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        |  n/a

* Excludes deprecated resources and fields
* Requires api-doc-parser 0.4
* Bumps dependecies
* Fixes tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "react-test-renderer": "^15.6.1"
   },
   "dependencies": {
-    "@api-platform/api-doc-parser": "^0.3.2",
+    "@api-platform/api-doc-parser": "^0.4",
     "admin-on-rest": "^1.4.1",
     "babel-runtime": "^6.23",
     "jsonld": "^0.4",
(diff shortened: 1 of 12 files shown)


mauchede:
@dunglas: Maybe, can we only remove deprecated fields from "defaultFields". Final users will have the ability to manage their own fields regardless of the default rule.

For example, in `admin/Edit.js`:

```js
const resolveProps = props => {
  const {options} = props;
  const {inputFactory: defaultInputFactory, resource} = options;
  const {
    editFields: customFields,
    editProps = {},
    writableFields: defaultFields,
  } = resource;
  const {options: {inputFactory: customInputFactory} = {}} = editProps;

  return {
    ...props,
    ...editProps,
    options: {
      ...options,
      fields: customFields || defaultFields.filter({ deprecated } => !deprecated),
      inputFactory: customInputFactory || defaultInputFactory,
    },
  };
};
```

dunglas:
@mauchede indeed, it's way better. Can you review the last commit please?

PR #96, opened 2018-06-02
Support embedded relations

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Currently, if the returned JSON-LD document contains [embedded relations](https://api-platform.com/docs/core/serialization/#embedding-relations), Admin isn't able to deal with it an throws an errors.

This PR let Admin find the `@id` of the nested document to retrieve it.

### DIFF
--- a/src/hydra/hydraClient.js
+++ b/src/hydra/hydraClient.js
@@ -27,9 +27,11 @@ class AORDocument {
    * @return {string}
    */
   toString() {
-    return this.id || super.toString();
+    return this.id;
   }
 }
+
+const aorDocumentsCache = new Map();
 
 /**
  * Transforms a JSON-LD document to an admin-on-rest compatible document
@@ -45,9 +47,15 @@ export const transformJsonLdDocumentToAORDocument = (
     return documents;
   }
 
-  documents = Array.isArray(documents)
-    ? Array.from(documents)
-    : new AORDocument(documents);
+  if (Array.isArray(documents)) {
+    documents = Array.from(documents);
+  } else if (documents['@id']) {
+    documents = new AORDocument(documents);
+    aorDocumentsCache[documents.toString()] = documents;
+  } else {
+    documents = Object.assign({}, documents);
+  }
+
   if (depth >= maxDepth) {
     return documents;
   }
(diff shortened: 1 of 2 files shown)


dunglas:
I just pushed another commit that let Admin reuse the data available in the nested document (if available) instead of sending an extra HTTP request to display the link to the relation. It dramatically improves performance.

dunglas:
I just pushed a less intrusive solution, less hackish solution that also fixes some bugs:

* Relations to other JSON-LD documents (to-one or to-many) are always replaced by the IRI itself. It preserves the default AOR behavior.
* Every replaced relation is also stored in a cache to prevent useless extra HTTP requests to this relation.

src/hydra/hydraClient.js
```diff
@@ -11,43 +11,89 @@ import isPlainObject from 'lodash.isplainobject';
...
-) => documents => {
-  if (!isPlainObject(documents) && !Array.isArray(documents)) {
-    return documents;
+  document,
+  clone = true,
+  addToCache = true,
+) => {
+  if (clone) {
+    // deep clone documents
+    document = JSON.parse(JSON.stringify(document));
```
soyuka: meh, I hope that this is never recursive or it'll break. you could use something like [fclone](https://github.com/soyuka/fclone), not sure if needed

src/hydra/hydraClient.js
```diff
@@ -11,43 +11,89 @@ import isPlainObject from 'lodash.isplainobject';
...
+      originId: obj.id,
+      id: obj['@id'],
+    });
+  }
+
+  /**
+   * @return {string}
+   */
+  toString() {
+    return `[object ${this.id}]`;
```
soyuka: we could also json.stringify, not sure how this is used.

dunglas: It’s just a hint when debugging

src/hydra/hydraClient.js
dunglas: It comes from a JSON, so cannot be recursive.

PR #97, opened 2018-06-05
Upgrade to React Admin

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

Not working yet: WIP.


src/hydra/hydraClient.js
```diff
@@ -6,12 +6,12 @@ import {
   GET_MANY_REFERENCE,
```
mauchede: Maybe can we rename this file `dataProvider.js`?

dunglas: This hasn't been done for React Admin built-in clients. I don't think it makes sense. (a restClient is a kind of data provider)

PR #99, opened 2018-06-07
bugfix: delete should not show invalid rest responses message

Simperfit:


src/hydra/hydraClient.js
```diff
@@ -302,7 +302,7 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
           .then(data => ({data, total: response.json['hydra:totalItems']}));
 
       case DELETE:
-        return Promise.resolve(() => ({data: {}}));
+        return Promise.resolve(response => (response.data: {}));
```
soyuka:
```
Promise.resolve(response => response.data)
```

src/hydra/hydraClient.js
```diff
@@ -302,7 +302,7 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
           .then(data => ({data, total: response.json['hydra:totalItems']}));
 
       case DELETE:
-        return Promise.resolve(() => ({data: {}}));
+        return Promise.resolve(response => (response.data));
```
soyuka: parenthesis not needed in the `(response.data)`

mauchede: Or you can use `destructuring`:

```js
({ data }) => data
```

k911: Just to notice: in DELETE_MANY you just do basically:

```return {data: {}};```

which fixes also this error :smile:

Simperfit: You can delete without using delete_many

Simperfit: @mauchede using destructuring generate more codes with yarn build than just this.

soyuka: @Simperfit is no js-hipster :cry:

k911:
Is there any chance that this fix will be finished and released soon? On admin application with enabled authorization, this error somehow causes further request to not have an authorization header which leads to user logout. When fix is applied everything works correctly.

Simperfit:
yarn fix has been applied, waiting for CI then merging this.

k911:
@Simperfit It seems it's still formatted unproperly.

k911:
Hi again @Simperfit @dunglas @mauchede, any news on this? I can submit PR that passes CI checks, if you want.

This bugfix is important to me, because apparently after deleting single element, admin won't send 'Authorization' headers in next requests and it makes user logged out, because of failed request.

dunglas:
@k911 it would be nice!

k911:
This can be closed, I think.

Simperfit:
Sorry I was on vacation ;)

PR #100, opened 2018-06-08
Filters implementation

jfthuillier:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | 
| Fixed tickets | #4
| License       | MIT
| Doc PR        | 

This PR on api-doc-parser must be merged before : https://github.com/api-platform/api-doc-parser/pull/33

### DIFF
--- a/src/fieldFactory.js
+++ b/src/fieldFactory.js
@@ -13,6 +13,16 @@ import {
 import React from 'react';
 import getReferenceNameField from './getReferenceNameField';
 
+export const isFieldSortable = (field, resource) => {
+  return (
+    resource.parameters.filter(parameter => parameter.variable === field.name)
+      .length > 0 &&
+    resource.parameters.filter(
+      parameter => parameter.variable === `order[${field.name}]`,
+    ).length > 0
+  );
+};
+
 export default (field, options) => {
   const props = {...field.fieldProps};
   if (field.field) {
@@ -33,6 +43,7 @@ export default (field, options) => {
           source={field.name}
           reference={field.reference.name}
           key={field.name}
+          sortable={isFieldSortable(field, options.resource)}
           {...props}
           allowEmpty>
           <ChipField source={getReferenceNameField(field.reference)} />
@@ -46,6 +57,7 @@ export default (field, options) => {
         source={field.name}
         reference={field.reference.name}
         key={field.name}
+        sortable={isFieldSortable(field, options.resource)}
         {...props}>
         <SingleFieldList>
           <ChipField source={referenceNameField} key={referenceNameField} />
@@ -56,10 +68,24 @@ export default (field, options) => {
 
   switch (field.id) {
     case 'http://schema.org/email':
-      return <EmailField key={field.name} source={field.name} {...props} />;
+      return (
+        <EmailField
+          key={field.name}
+          source={field.name}
+          sortable={isFieldSortable(field, options.resource)}
+          {...props}
+        />
+      );
 
     case 'http://schema.org/url':
-      return <UrlField key={field.name} source={field.name} {...props} />;
+      return (
+        <UrlField
+          key={field.name}
+          source={field.name}
+          sortable={isFieldSortable(field, options.resource)}
+          {...props}
+        />
+      );
 
     default:
     // Do nothing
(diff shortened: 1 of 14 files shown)


src/parameterFactory.js
```diff
@@ -0,0 +1,115 @@
...
+    return 'between';
+  }
+
+  return guessTypeFromResource(parameter, fields);
+}
+
+function guessTypeFromResource(parameter, fields) {
+  const type = fields
+    .filter(field => {
+      return field['name'] === parameter.variable;
```
arnaudoisel: Fixed (parameter.property => parameter.variable) so that range and nullableboolean fields can be generated.

src/fieldFactory.js
```diff
@@ -26,13 +36,14 @@ export default (field, options) => {
     );
   }
 
-  if (null !== field.reference) {
+  if (null != field.reference) {
```
arnaudoisel: Changed to check for undefined

src/inputFactory.js
```diff
@@ -28,7 +28,7 @@ export default (field, options) => {
 
   if (!props.validate && field.required) props.validate = [required()];
 
-  if (null !== field.reference) {
+  if (null != field.reference) {
```
arnaudoisel: Changed to check for undefined

src/parametersFactory.js
```diff
@@ -0,0 +1,16 @@
+import parameterFactory from './parameterFactory';
+import parametersDuplicatesFilter from './parametersDuplicatesFilter';
+
+export function filterOutDuplicatedVariablesEndingWithBrackets(array) {
+  return parametersDuplicatesFilter(array, 'variable', '[]');
```
arnaudoisel: Added to filter out duplicated filters.

src/fieldFactory.js
jfthuillier: I've dropped this commit, the `field.reference` should be null of have a value, there must be something wrong if it is undefined. Can you give steps to reproduce ?

src/parametersFactory.js
jfthuillier: I've dropped this commit, IMO your solution is overkill. What i had in mind was just a regex : https://github.com/api-platform/admin/blob/e81e8537b9fd2338458f107183e87c4dae71821e/src/parameterFactory.js#L63

Maybe you discuss about this with @dunglas and there is a reason i am missing. Anyway, i saved your work on another branch, don't worry : https://github.com/jfthuillier/admin/commits/filters-backup.

src/parameterFactory.js
jfthuillier: Agreed, i renamed from `property` to `variable` in another commit and i forgot this line

src/fieldFactory.js
arnaudoisel: Yes you are right, I made this change while I was missing your last commits.

src/parametersFactory.js
arnaudoisel: It was to keep properties in the case where the API returns a variable ending with square brackets and not the same variable without square brackets. For exemple, to keep intact :
`{ variable: "foo[]" }`
But I am not sure if this case is possible with API Platform. So it might be overkill indeed. I will discuss this with @dunglas

src/hydra/hydraClient.js
```diff
@@ -201,9 +201,19 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
...
-          );
+          Object.keys(params.filter).forEach(key => {
+            const filterValue = params.filter[key];
+            if (isPlainObject(filterValue)) {
+              Object.keys(filterValue).forEach(subKey => {
+                collectionUrl.searchParams.set(
+                  `${key}[${subKey}]`,
+                  filterValue[subKey],
+                );
+              });
```
dunglas: You could add a `return` here to avoid the next `if` and reduce a bit the cyclomatic complexity

src/parameterFactory.js
```diff
@@ -0,0 +1,124 @@
...
+  DateInput,
+  EmailInput,
+  NullableBooleanInput,
+  NumberInput,
+  TextInput,
+} from 'react-admin';
+import React from 'react';
+
+function guessType(parameter, fields) {
+  if (parameter.variable.match(/.*\[exists\]/i) !== null) {
```
dunglas: The `return` could be on the same line for consistency (IIRC, we do that evrywhere), same for similar one-line constructs

src/parameterFactory.js
```diff
@@ -0,0 +1,124 @@
+import {
+  DateInput,
+  EmailInput,
+  NullableBooleanInput,
+  NumberInput,
+  TextInput,
+} from 'react-admin';
+import React from 'react';
+
+function guessType(parameter, fields) {
```
dunglas: I'm not a fond of this function because it's hardcoded for specific API Platform values, couldn't we use the Hydra metadata instead? The type is provided.

src/parameterFactory.js
```diff
@@ -0,0 +1,124 @@
...
+      return 'text';
+    });
+
+  return type.length > 0 ? type[0] : 'text';
+}
+
+export default (parameter, fields, options) => {
+  if (
+    // List filters are discarded because there is no built-in filter component in react-admin that can handle this case.
+    parameter.variable.match(/.*\[\]/i) ||
```
dunglas: Maybe could we introduce a `api-platform` flag allowing to disable all these "vendor specific" tweaks?

dunglas:
Thank you very much for this great new feature @jfthuillier and @arnaudoisel!

PR #107, opened 2018-07-23
Do not alter relations when GET_ONE

jfthuillier:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | 

When a `GET_ONE` is performed the `hydraClient` automatically replace related entities by their `id`.
_Example :_ 
If the API return is : 
```
{
    "fieldLvl1":"dzadza",
    "linkedEntity":{
        "id": 1,
        "fieldLvl2":"dzadza",
    }
}
```

The client will return : 
```
{
    "fieldLvl1":"dzadza",
    "linkedEntity": 1
}
```

This object is then passed to the Edit component as `record` props. So if an input with `linkedEntity.fieldLvl2` is declared, it will not get his expected value.

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -12,6 +12,18 @@ import isPlainObject from 'lodash.isplainobject';
 import fetchHydra from './fetchHydra';
 import parseHydraDocumentation from '@api-platform/api-doc-parser/lib/hydra/parseHydraDocumentation';
 
+const isIri = (key, value) => {
+  if (typeof value !== 'string') {
+    return false;
+  }
+
+  if (key === 'id' || key === '@id') {
+    return false;
+  }
+
+  return value.match(/\/[a-z]+\/[0-9]+/g);
+};
+
 class ReactAdminDocument {
   constructor(obj) {
     Object.assign(this, obj, {
@@ -41,6 +53,8 @@ const reactAdminDocumentsCache = new Map();
  *
  * @param {Object} document
  * @param {bool} clone
+ * @param {bool} addToCache
+ * @param {bool} doNoAlterRelations
  *
  * @return {ReactAdminDocument}
  */
@@ -48,6 +62,7 @@ export const transformJsonLdDocumentToReactAdminDocument = (
   document,
   clone = true,
   addToCache = true,
+  doNoAlterRelations = false,
 ) => {
   if (clone) {
     // deep clone documents
@@ -72,9 +87,17 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           false,
         );
       }
-      document[key] = document[key]['@id'];
+
+      document[key] =
+        doNoAlterRelations || !document[key]['@id']
+          ? document[key]
+          : document[key]['@id'];
 
       return;
+    }
+
+    if (doNoAlterRelations && isIri(key, document[key])) {
+      document[key] = {'@id': document[key], id: 1};
     }
 
     // to-many
@@ -91,8 +114,10 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           ] = transformJsonLdDocumentToReactAdminDocument(obj, false, false);
         }
 
-        return obj['@id'];
+        return doNoAlterRelations ? obj : {'@id': obj['@id'], id: 1};
       });
+
+      document[`${key}_ids`] = document[key].map(obj => obj['@id']);
     }
   });
 
@@ -331,6 +356,18 @@ export default (
 
       case DELETE:
         return Promise.resolve({data: {id: null}});
+
+      case GET_ONE:
+        return Promise.resolve(
+          transformJsonLdDocumentToReactAdminDocument(
+            response.json,
+            true,
+            true,
+            true,
+          ),
+        )
+          .then(data => convertHydraDataToReactAdminData(resource, data))
+          .then(data => ({data}));
 
       default:
         return Promise.resolve(
--- a/src/InputGuesser.js
+++ b/src/InputGuesser.js
@@ -38,8 +38,11 @@ export const InputGuesserComponent = ({fields, resourceSchema, ...props}) => {
           reference={field.reference.name}
           source={field.name}
           validate={validate}
-          {...props}
-          allowEmpty>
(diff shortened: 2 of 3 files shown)


src/hydra/hydraClient.js
```diff
@@ -72,7 +75,12 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           false,
         );
       }
-      document[key] = document[key]['@id'];
+
+      if (doNoAlterRelations) {
```
ronfroy:
```if (!doNoAlterRelations) {
        document[key] = document[key]['@id'];
}```

no need a else

Comment (jfthuillier) [reply src/hydra/hydraClient.js]: You're right, thank you :+1:

File: src/hydra/hydraClient.js [hunk 2]
```diff
@@ -91,7 +99,11 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           ] = transformJsonLdDocumentToReactAdminDocument(obj, false, false);
         }
 
-        return obj['@id'];
+        if (doNoAlterRelations) {
```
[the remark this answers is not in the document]
jfthuillier: Done

src/hydra/hydraClient.js
```diff
@@ -91,7 +95,11 @@ export const transformJsonLdDocumentToReactAdminDocument = (
...
```
dunglas:
```suggestion
        return doNoAlterRelations ? obj : obj['@id'];
```

src/hydra/hydraClient.js
```diff
@@ -72,7 +75,8 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           false,
         );
       }
-      document[key] = document[key]['@id'];
+
+      document[key] = doNoAlterRelations ? document[key] : document[key]['@id'];
```
dunglas:
```suggestion
      document[key] = doNoAlterRelations || !document[key]['@id'] ? document[key] : document[key]['@id'];
```

src/hydra/hydraClient.js
jfthuillier: Copy/paste has been done :laughing:

arnaudoisel:
I don't manage to receive a response of this type from API Platform :
```
{
    "fieldLvl1":"dzadza",
    "linkedEntity":{
        "id": 1,
        "fieldLvl2":"dzadza",
    }
}
```
```
I always get a response of type :
{
    "fieldLvl1":"dzadza",
    "linkedEntity": 1
}
```

Could you provide an exemple of API Entities that could be used to generate such a hydra response please ?

comxd:
@jfthuillier any news about this?

jfthuillier:
I just pushed a new commit with what could be a way to fix the side effect on `ReferenceInput`. It could work the same way with the `ReferenceArrayInput` but i'm not really satisfied with the `isIri` function.
WDYT ?

comxd:
It working good for all cases in my tests.

comxd:
@dunglas Can you review this PR please?

src/hydra/hydraClient.js
```diff
@@ -11,6 +11,18 @@ import {
...
+const isIri = (key, value) => {
+  if (typeof value !== 'string') {
+    return false;
+  }
+
+  if (key === 'id' || key === '@id') {
+    return false;
+  }
+
+  return value.match(/\/[a-z]+\/[0-9]+/g);
```
soyuka: not sure why we need this regular expression?

src/hydra/hydraClient.test.js
```diff
@@ -73,6 +73,19 @@ describe('map a json-ld document to an admin on rest compatible document', () =>
...
+      true,
+      true,
+      true,
+    );
+
+    test('has not transform arrays of embedded documents to their IRIs', () => {
+      expect(reactAdminDocument.comment[0]).not.toBe('/comments/1');
+    });
+  });
 });
```
soyuka: can we have a test where we go through the path where `isIri(key, document[key])` is falsy? Or else, I'm not sure if this function is really useful as used only once.

comxd:
Is this problem still relevant with version 1.0.0?

src/hydra/dataProvider.js
```diff
@@ -12,6 +12,18 @@ import isPlainObject from 'lodash.isplainobject';
...
+const isIri = (key, value) => {
+  if (typeof value !== 'string') {
+    return false;
+  }
+
+  if (key === 'id' || key === '@id') {
+    return false;
+  }
+
+  return value.match(/\/[a-z]+\/[0-9]+/g);
```
soyuka: I still don't get how this is a match for an IRI ^^.

src/hydra/dataProvider.js
```diff
@@ -12,6 +12,18 @@ import isPlainObject from 'lodash.isplainobject';
 import fetchHydra from './fetchHydra';
 import parseHydraDocumentation from '@api-platform/api-doc-parser/lib/hydra/parseHydraDocumentation';
 
+const isIri = (key, value) => {
+  if (typeof value !== 'string') {
+    return false;
+  }
+
+  if (key === 'id' || key === '@id') {
+    return false;
```
soyuka: we skip these? Because they're probably iris

mleczakm:
Any idea when this will be fixed? This bug makes custom serialization groups totally unusable with resources, works only with non-resource objects.

alanpoulain:
I'm not sure if this feature is useful. Why is it needed? You can show the related embedded fields by using a reference like described here: https://api-platform.com/docs/admin/handling-relations/#display-a-field-of-an-embedded-relation.

mleczakm:
@alanpoulain How about non-form context, for example inside validators etc. where you are not able to fetch `author` with second call, like it will be done with `ReferenceField`? It is not feature, it is bugfix - APAdmin is removing data different than iri from incoming response.

alanpoulain:
Replaced by https://github.com/api-platform/admin/pull/299.

Issue #118, opened 2018-09-25
Unexpected 404 when API is exposed as the /api/ path


dloubere:
Having installed the API this way => https://api-platform.com/docs/distribution#using-symfony-flex-and-composer-advanced-users and the admin following https://api-platform.com/docs/admin/getting-started, I'm getting 404 because the admin is sometimes calling `/api/api/...` routes:

```
2018-09-25T11:40:00+02:00 [info] Matched route "api_entrypoint".
2018-09-25T11:40:00+02:00 [info] Populated the TokenStorage with an anonymous Token.
2018-09-25T11:40:00+02:00 [info] User Deprecated: Doctrine\Common\ClassLoader is deprecated.
[Tue Sep 25 11:40:00 2018] 127.0.0.1:48694 [200]: /api
2018-09-25T11:40:00+02:00 [info] Matched route "api_doc".
2018-09-25T11:40:00+02:00 [info] Populated the TokenStorage with an anonymous Token.
2018-09-25T11:40:00+02:00 [info] User Deprecated: Doctrine\Common\ClassLoader is deprecated.
[Tue Sep 25 11:40:00 2018] 127.0.0.1:48698 [200]: /api/docs.jsonld
2018-09-25T11:40:00+02:00 [info] Matched route "api_jsonld_context".
2018-09-25T11:40:00+02:00 [info] Populated the TokenStorage with an anonymous Token.
[Tue Sep 25 11:40:00 2018] 127.0.0.1:48700 [200]: /api/contexts/Entrypoint
2018-09-25T11:40:00+02:00 [info] Matched route "api_reviews_get_collection".
2018-09-25T11:40:00+02:00 [info] Populated the TokenStorage with an anonymous Token.
2018-09-25T11:40:00+02:00 [info] User Deprecated: Doctrine\Common\ClassLoader is deprecated.
2018-09-25T11:40:00+02:00 [debug] SELECT DISTINCT r0_.id AS id_0, r0_.id AS id_1 FROM review r0_ ORDER BY r0_.id ASC LIMIT 30
2018-09-25T11:40:00+02:00 [debug] SELECT r0_.id AS id_0, r0_.rating AS rating_1, r0_.body AS body_2, r0_.author AS author_3, r0_.publication_date AS publication_date_4, r0_.book_id AS book_id_5 FROM review r0_ WHERE r0_.id IN (?) ORDER BY r0_.id ASC
[...]

PR #124, opened 2018-10-25
Fix hydra client with prefixed API URL

ginifizz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | api-platform/api-platform#584, api-platform/admin#118
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/hydra/hydraClient.js
+++ b/src/hydra/hydraClient.js
@@ -168,9 +168,8 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
    * @returns {Object}
    */
   const convertReactAdminRequestToHydraRequest = (type, resource, params) => {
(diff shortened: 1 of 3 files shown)


dunglas:
Thank you very much for fixing this @ginifizz!!

johnstyle:
This does not seem to work with a relative entrypoint `entrypoint='/api'` ?
I have this error `Failed to construct 'URL': Invalid URL` with this fix.

dunglas:
To support relative entrypoint, we should:

* test if it relative or not
* append the relative to `window.location.origin` if it is relative

Would someone want to work on a Pull Request?

PR #126, opened 2018-10-31
Fix relative entrypoint on hydra client

ginifizz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #124
| License       | MIT
| Doc PR        | api-platform/doc#1234

### DIFF
--- a/src/hydra/hydraClient.js
+++ b/src/hydra/hydraClient.js
@@ -9,7 +9,6 @@ import {
   UPDATE,
 } from 'react-admin';
 import isPlainObject from 'lodash.isplainobject';
-import qs from 'qs';
 import fetchHydra from './fetchHydra';
 
 class ReactAdminDocument {
@@ -169,6 +168,9 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
    * @returns {Object}
    */
   const convertReactAdminRequestToHydraRequest = (type, resource, params) => {
+    const collectionUrl = new URL(`${entrypoint}/${resource}`);
+    const itemUrl = new URL(params.id, entrypoint);
+
     switch (type) {
       case CREATE:
         return transformReactAdminDataToRequestBody(resource, params.data).then(
@@ -177,7 +179,7 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
               body,
               method: 'POST',
             },
-            url: `${entrypoint}/${resource}`,
+            url: collectionUrl,
           }),
         );
 
@@ -186,7 +188,7 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
           options: {
             method: 'DELETE',
           },
-          url: entrypoint + params.id,
+          url: itemUrl,
         });
 
       case GET_LIST: {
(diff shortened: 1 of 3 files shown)


src/hydra/hydraClient.js
```diff
@@ -168,8 +168,9 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
    * @returns {Object}
    */
   const convertReactAdminRequestToHydraRequest = (type, resource, params) => {
-    const collectionUrl = new URL(`${entrypoint}/${resource}`);
-    const itemUrl = new URL(params.id, entrypoint);
+    const entrypointUrl = new URL(entrypoint, window.location.origin);
```
dunglas: We could even use `window.location.href` to support relative paths (as in your test)

arnaudoisel: I don't see any difference between window.location.origin and window.location.href, except that window.location.href would include potential anchors (https://domain#foo), but I think they are treated the same way by the URL class.

src/hydra/hydraClient.js
```diff
@@ -168,8 +168,9 @@ export default ({entrypoint, resources = []}, httpClient = fetchHydra) => {
...
+    const collectionUrl = new URL(`${entrypoint}/${resource}`, entrypointUrl);
```
arnaudoisel: Could it be :
```suggestion
    const collectionUrl = new URL(resource, entrypointUrl);
```

src/hydra/hydraClient.js
arnaudoisel: I don't see any difference between window.location.origin and window.location.href, except that window.location.href would include potential anchors (https://domain#foo), but I think they are treated the same way by the URL class.

src/hydra/hydraClient.js
arnaudoisel: Could it be :
```suggestion
    const collectionUrl = new URL(resource, entrypointUrl);
```

src/hydra/hydraClient.js
dunglas:
```
> window.location.href
< "https://github.com/api-platform/admin/pull/126"
> window.location.origin
< "https://github.com"
```

With `window.location.href`, relative path will work.

dunglas:
```
> new URL('foo/bar', window.location.origin)
< URL {href: "https://github.com/foo/bar", origin: "https://github.com", protocol: "https:", username: "", password: "", …}
> new URL('foo/bar', window.location.href)
< URL {href: "https://github.com/api-platform/admin/pull/foo/bar", origin: "https://github.com", protocol: "https:", username: "", password: "", …}
```

arnaudoisel: Right, sorry, my bad. I was only testing in the case where the base URL is the origin...

```
> new URL('foo/bar', (new URL('https://github.com/api-platform')).origin)
< URL { href: "https://github.com/foo/bar" }
> new URL('foo/bar', (new URL('https://github.com/api-platform')).href)
< URL { href: "https://github.com/foo/bar" }
```

dunglas: Wow, it looks like a browser bug to me. The origin should never include the path.

arnaudoisel: It doesn't, my example is just the special case where the path length is 1. In that case the base URL is the origin.

Issue #122, opened 2018-10-11
Cannot manage datetime or array fields (and probably others)


wildcard27:
I have a field that is of type: array. I am trying to save data in this field but the admin component keeps giving me a The type of the "acceptedMeasurements" attribute must be "array", "string" given. error (obviously from Symfony).

Here is the relevant code:
```
/**
 * @ORM\Column(type="array", nullable=true)
 */
private $acceptedMeasurements;


/**
 * @return null|array
 */
public function getAcceptedMeasurements(): ?array
{
    return $this->acceptedMeasurements;
}

/**
 * @param null|array $acceptedMeasurements
 *
 * @return Exercise
 */
public function setAcceptedMeasurements(?array $acceptedMeasurements): self
{
    $this->acceptedMeasurements = $acceptedMeasurements;

    return $this;
}
```

Swagger seems to have this correct:
```
{
  "active": true,
  "name": "string",
  "user": "string",
  "acceptedMeasurements": [
    "string"
  ]
}
```

In admin, I have tried inputting:
```
[test]
['test']
["test"]
{"test"}
a:1:{i:0;s:4:"test";}
```
and probably a couple more but nothing works.

To add to this, you cannot add a time to datetime fields. I assume there are other field types that don't work, too but haven't found them yet.

krcrawford:
I was able to get this working, assuming the following condition: 

- The list action will parse the array into a string, this is the default behavior (so no customization needed here unless you want a bunch of chips. in which case the following technique does not work)

```
...
import { ArrayInput, SimpleFormIterator, TextInput } from 'ra-ui-materialui';
...
const user = api.resources.find(({ name }) => 'users' === name);
const roles = user.fields.find(({ name }) => 'roles' === name);
roles.input = props => (
    <ArrayInput source="roles">
        <SimpleFormIterator>
            <TextInput/>
        </SimpleFormIterator>
    </ArrayInput>
);
...
```

dunglas:
Thanks for the feedback! Would you mind to open a Pull Request to support this natively @krcrawford?

krcrawford:
@dunglas PR submitted

krcrawford:
@dunglas possibly close the issue?

dunglas:
Thanks @krcrawford

PR #134, opened 2018-11-16
Adds tests and native support for unserialized unrelated arrays

krcrawford:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #122
| License       | MIT
| Doc PR        | api-platform/doc#664

This PR supports a simple strategy to edit serialized array inputs from DB entities.
While it may not currently address more complicated array structures, this solution
supports simple one-dimensional, keyless(numerically indexed), arrays


krcrawford:
This may also require an update to core as it seems a serialized array gets transported as a string. For now, the core wont send a range like: `range: 'http://www.w3.org/2001/XMLSchema#array` under these circumstances

Issue #132, opened 2018-11-16
Error when deleting a resource


trsteel88:
When deleting a resource I am getting the following error in console:

```
The response to 'DELETE' must be like { data: { id: 123, ... } }, but the received data does not have an 'id' key. The dataProvider is probably wrong for 'DELETE'
```

The record is deleted. But the UI has an error. Is there an issue at the moment?

krcrawford:
Getting the same issue here

baudev:
Also.
There is also the equivalent when deleting from the list:

```
The response to 'DELETE_MANY' must be like { data : [...] }, but the received data is not an array. The dataProvider is probably wrong for 'DELETE_MANY'
```

PR #140, opened 2018-12-20
Quick fix for issue #132 : Error when deleting a resource

arnaudoisel:
+ react-admin should be fixed instead, for long-term fix.
+ react-admin should not expect a content in the response after a delete or delete_many

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #132
| License       | MIT
| Doc PR        |


dunglas:
Thank you very much @arnaudoisel

Issue #129, opened 2018-11-08
Admin platform does not handle server side errors


desmax:
When I edit some entity and press save it goes to screen with list of entities without checking server side errors. When it actually sends the request (and receives 400 code with list of violations) it just shows "Bad request" message (and for some reason sends GET request to /contexts/ConstraintViolationList which fails with 401 code). 

What can I do to make it show server errors? (for example "email already taken")

trsteel88:
I am also interested in how to do this.

dunglas:
It looks like a regression. Can anyone check what happens?

thomas-obernberger:
what do you mean by regression?

looks like it happens here
node_modules/@api-platform/admin/src/hydra/fetchHydra.js:31-34:
```
promises
  .expand(data.body, {
    base: getDocumentationUrlFromHeaders(data.response.headers),
})
```
---

this promise is rejected, jsond tries to access `https://hostname/contexts/ConstraintViolationList` -> but this returns an error `401 Unauthorized`, because its not using the initial set headers (with an Authorization header)
```
data.body:
  @context: "/contexts/ConstraintViolationList"
  @type: "ConstraintViolationList"
  hydra:description: "qualifications: Dieser Wert sollte nicht leer sein."
  hydra:title: "An error occurred"
  violations: [{…}]

base:
  https://hostname/docs.jsonld
```

---
then i tried to override XMLHttpRequest to force add the Authorization header
```
XMLHttpRequest.prototype.origOpen = XMLHttpRequest.prototype.open;
XMLHttpRequest.prototype.open   = function () {
    this.origOpen.apply(this, arguments);
    this.setRequestHeader('Authorization', `Bearer ${window.localStorage.getItem('token')}`);
};
```
(i know this is insecure code - i just implemented it for debugging..)
the request to `https://hostname/contexts/ConstraintViolationList` is working now and the promise succeeds.
.. then on line 36 the promise will be rejected with following data
```
[json[0]['http://www.w3.org/ns/hydra/core#description'][0]['@value'], status] contains:
0: "qualifications: Dieser Wert sollte nicht leer sein."
1: 400
```
-> then at line 46 the excection should be an instance of `HttpError` (`if (e instanceof HttpError) {`)
this condition is not working and instead of the previous built error message - the http error message is returned
```
content of e:
qualifications: Dieser Wert sollte nicht leer sein.
    at new HttpError (HttpError.js:56)
    at fetchHydra.js:64
```
---
i hope this information is helpful
[...]

PR #141, opened 2018-12-27
Fix issue : Admin platform does not handle server side errors #129

arnaudoisel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #129 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/hydra/fetchHydra.js
+++ b/src/hydra/fetchHydra.js
@@ -43,7 +43,7 @@ export default (url, options = {}) => {
           );
         })
         .catch(e => {
-          if (e instanceof HttpError) {
+          if (e.hasOwnProperty('body')) {
             return Promise.reject(e);
           }
 


arnaudoisel:
'react-admin' moved from Babel transpiler to Typescript transpiler (https://github.com/marmelab/react-admin/pull/2426).
And there is a known issue with 'instanceof' and Error subclasses (https://github.com/Microsoft/TypeScript/issues/13965).
So since v2.4.x of 'react-admin', it is not possible anymore to expect an instance of a subclass of Error to be an instanceof this subclass.

FYI: with Babel there used to be a similar problem but there was a plugin to fix this (https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend). Such a plugin does not exist with Typescript transpiler.

dunglas:
Great investigation! Thanks for the fix @arnaudoisel

PR #155, opened 2019-02-24
Add a filter function to remove fields from the list pages

bakulf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I need to remove some fields from the List pages. With this commit I introduce a filter function which can be used to select what needs be shown and what hidden from the List component.

The way I use the new function is something like this:
```

const hiddenFromList = [
  { entity: "people", fields: [ "address", "friends", ]},
  { entity: "books", fields: [ "reviews", ]},
];

const listFieldFilter = (resource, field) => {
  const hiddenResource = hiddenFromList.find(hidden => hidden.entity === resource.name);
  return !hiddenResource || !hiddenResource.fields.includes(field.name);
};

export default props => (
    <HydraAdmin
       ... // here the apiDocumentationParser, authProvider, entrypoint, etc
        listFieldFilter={listFieldFilter}
    />
);
```

### DIFF
--- a/src/AdminBuilder.js
+++ b/src/AdminBuilder.js
@@ -30,6 +30,7 @@ const AdminBuilder = props => {
     parameterFactory,
     title = api.title,
     resources = api.resources.filter(({deprecated}) => !deprecated),
+    listFieldFilter,
   } = props;
 
   return (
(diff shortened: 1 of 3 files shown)


dunglas:
Thanks @bakulf

PR #172, opened 2019-04-25
Fix Hydra Admin promises

### LINKED ISSUE #171: Error when no resource return by api doc parser
The following code in [`HydraAdmin`](https://github.com/api-platform/admin/blob/master/src/hydra/HydraAdmin.js#L37) lead to an error breaking the admin : 
``` js
const promises = api.resources.forEach(resource =>
            promises.push(resource.getParameters()),
          );

          return Promise.all(promises).then(responses => {
            api.resources.forEach((resource, index) => {
              resource.parameters = responses[index];
            });

            return {
              api,
              customRoutes,
              hasError: false,
              loaded: true,
            };
          });
```

if `api.resources` is empty (ex: `401` when not logged), `promises` is `undefined` and `Promise.all(promises)` throw an error.

Introduced with : #150

jordanaubert:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #171 
| License       | MIT

### DIFF
--- a/src/hydra/HydraAdmin.js
+++ b/src/hydra/HydraAdmin.js
@@ -34,8 +34,8 @@ export default class extends Component {
       .apiDocumentationParser(this.props.entrypoint)
       .then(
         ({api, customRoutes = []}) => {
-          const promises = api.resources.forEach(resource =>
-            promises.push(resource.getParameters()),
+          const promises = api.resources.map(resource =>
+            resource.getParameters(),
           );
 
           return Promise.all(promises).then(responses => {


src/hydra/HydraAdmin.js
```diff
@@ -34,8 +34,8 @@ export default class extends Component {
       .apiDocumentationParser(this.props.entrypoint)
       .then(
         ({api, customRoutes = []}) => {
-          const promises = api.resources.forEach(resource =>
-            promises.push(resource.getParameters()),
+          const promises = api.resources.map(resource =>
+            resource.getParameters(),
```
soyuka:
```suggestion
            resource.getParameters()
```

jordanaubert: The linter seems to disagree with you 😀

alanpoulain:
Thank you @jordanaubert.

Issue #214, opened 2019-09-18
Submitting the edit form doesn't work


dunglas:
When clicking on the edit button, the loader is displayed and the page hangs.

PR #217, opened 2019-09-18
Fix edit and delete buttons

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #214
| License       | MIT
| Doc PR        | n/a

Setting `undoable=false` fixes the issue... but disables the nice "undo" feature.

The problem seems to be a race condition between the `LIST` and the `UPDATE` operations. Whe `undoable` is enabled:

1. the `LIST` request is sent
2. The `UPDATE` request is sent with a delay
3. The result of `LIST` doesn't dispatch the function wrapped in `<Query>` as it should, and the loading screen never disappears.

@fzaninotto do you have an idea of how we can fix this?

Thanks to @mauchede for finding the underlying issue.

### DIFF
--- a/src/EditGuesser.js
+++ b/src/EditGuesser.js
@@ -6,10 +6,10 @@ import WithReactAdminQuery from './withReactAdminQuery';
 
 // useful for testing (we don't need Query)
 const EditGuesserComponent = props => {
-  const {children, fields} = props;
+  const {children, fields, resourceSchema, ...filteredProps} = props;
 
   return (
-    <Edit {...props}>
+    <Edit {...filteredProps} undoable={false}>
       <SimpleForm>
         {children}
         {fields.map(field => (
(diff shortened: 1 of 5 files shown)


fzaninotto:
Sorry I don't understand the issue. Could you please give me more context? Also, as there is no such bug in react-admin core, I believe the problem lies in the guessers.

dunglas:
WIthout setting `undoable` to `false`, when you submit the edit form, or click on the "delete" button, the app is blocked on the loading screen. According to our finding, the problem is related to the fact that several queries are send in parallel (using the `<Query>` component (first the new INTROSPECT query to display the list, then UPDATE, because it's delayed for the undo feature). In this order, the result of the `INTROSPECT` query is never dispatched.

It's probably not a bug in ra-core. We'll try to investigate deeper.

PR #224, opened 2019-09-19
Major refactoring, and major config improvement

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/872

See https://github.com/api-platform/docs/pull/872 for the detailed documentation.


dunglas:
Example added in the demo: https://github.com/api-platform/demo/pull/128

src/AdminGuesser.js
```diff
@@ -3,10 +3,14 @@ import PropTypes from 'prop-types';
 import {Admin, Loading, Query, TranslationProvider} from 'react-admin';
 import {createMuiTheme} from '@material-ui/core/styles';
 import ResourceGuesser from './ResourceGuesser';
-import {existsAsChild} from './docsUtils';
 
-export const render = (
-  {blacklist = [], children, ...props},
+/**
+ * AdminGuesserComponent automatically renders and `<Admin>` component for resources exposed by a web API documented with Hydra, OpenAPI or any other format supported by `@api-platform/api-doc-parser`.
```
chalasr: typo `renders and` => `renders an`

src/AdminGuesser.js
```diff
@@ -3,10 +3,14 @@ import PropTypes from 'prop-types';
...
+ * If children components are passed (usually `<ResourceGuesser>` or `<Resource>` components, but it can be any other React component), they are rendered in the given order.
```
chalasr: I'd say `child components` (see e.g. https://developer.mozilla.org/fr/docs/Web/API/Node/childNodes)

src/AdminGuesser.js
```diff
@@ -24,25 +28,19 @@ export const render = (
...
-    .forEach(resource =>
-      childrenAsArray.push(
-        <ResourceGuesser name={resource.name} key={resource.name} />,
-      ),
-    );
+  if (!children) {
+    children = data.resources
+      .filter(r => !r.deprecated)
+      .map(r => <ResourceGuesser name={r.name} key={r.name} />);
+  }
```
jfthuillier: Is it on purpose to not handle a control boolean like in IntrospectQuery to include deprecated resources ?

src/ListGuesser.js
```diff
@@ -33,18 +33,21 @@ class ListGuesserComponent extends React.Component {
 
   render() {
     const {resourceSchema, fields, ...props} = this.props;
-    const children = Children.toArray(props.children);
+
+    if (!props.children) {
```
jfthuillier: Any reason to not destructure children, like in other guessers ?

dunglas: Because it's not a functional component. Deconstructing to a `const` is useless cause we have to override its value.

src/AdminGuesser.js
dunglas: Control props added. To be documented.

PR #240, opened 2019-10-12
Allow to disable filters

tienvx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | NA
| License       | MIT
| Doc PR        | NA

1. This change allow us to disable filters:
```
<ListGuesser {...props}  filters={null}>
    <FieldGuesser source='test' />
  </ListGuesser>
```
2. We can't simply ommit `filters` property, because it has default value on `ListGuesser`
3. `filters` property is not required for `List`

### DIFF
--- a/src/ListGuesser.js
+++ b/src/ListGuesser.js
@@ -63,7 +63,7 @@ const ListGuesser = props => (
 ListGuesser.propTypes = {
   children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
   resource: PropTypes.string.isRequired,
-  filters: PropTypes.object.isRequired,
+  filters: PropTypes.element,
 };
 
 ListGuesser.defaultProps = {


src/ListGuesser.js
```diff
@@ -63,7 +63,6 @@ const ListGuesser = props => (
 ListGuesser.propTypes = {
   children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
   resource: PropTypes.string.isRequired,
-  filters: PropTypes.object.isRequired,
```
dunglas: `filters: PropTypes.object` then right?
Also, maybe passing `false` instead of  `null` would be more explicit?

dunglas: Ok for `null`, and good catch for the `PropTypes.element`! Could you change it then? Thanks!

dunglas:
Thanks a lot!

Issue #238, opened 2019-10-10
Upload of files as multipart fails without resource.encodeData in dataProvider


rkcreation:
Hi,

Since commit bb6c8918e50ddd34e7cc9341865d518a601e1820, I can't upload files as multipart. I follow API Platform docs and use `resource.encodeData` to make a FormData object before sending to API server, and after upgrading uploads don't work anymore.

Is there a reason for removing use of `resource.encodeData` in `transformReactAdminDataToRequestBody` ?
Any idea to achieve upload otherwise ?

API Platform docs says about `/media_objects`:
```This endpoint accepts standard multipart/form-data-encoded data, but not JSON data.```
https://api-platform.com/docs/core/file-upload/#making-a-request-to-the-media_objects-endpoint

Thanks !

PR #246, opened 2019-10-17
restore resource.encodeData

rkcreation:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #238 
| License       | MIT
| Doc PR        | 

Restore usage of resource.encodeData to avoid breaking existing apps (at least for those who follow API Platform examples).

Usage of encodeData is needed to send files to API which expects multipart FormData.

Really blocking to update v0.6.3 to v1.x.x

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -170,7 +170,9 @@ export default (
     }
 
     return convertReactAdminDataToHydraData(resource, data).then(data =>
-      JSON.stringify(data),
+      undefined === resource.encodeData
+        ? JSON.stringify(data)
+        : resource.encodeData(data),
     );
   };
 


src/hydra/dataProvider.js
```diff
@@ -170,7 +170,9 @@ export default (
     }
 
     return convertReactAdminDataToHydraData(resource, data).then(data =>
-      JSON.stringify(data),
+      undefined === resource.encodeData
+        ? JSON.stringify(data)
+        : resource.encodeData(data),
```
soyuka: could you add a test? I'd do the opposite btw:

```
resource.encodeData ? resource.encodeData(data) : JSON.stringify(data)
```

dunglas:
I don't think that it's the way recommended by React Admin: https://github.com/marmelab/react-admin/blob/master/docs/Inputs.md#fileinput

The goal of v1 was to get rid of all these custom, non-idiomatic stuffs to delegate to React Admin. Yes it introduces BC breaks, but it was intended. I would like to not re-introduce such hacks.

rkcreation:
@dunglas I agree with this, but for bypassing the JSON stringify of the hydra dataProvider it needs lot of code comparing with this resource.encodeData.

The example provided by react-admin converts a property to base64, but then it follows normal dataProvider way (requestHandler is the return of dataProvider init) :
```
               transformedNewPictures => requestHandler(type, resource, {
                    ...params,
                    data: {
                        ...params.data,
                        pictures: [...transformedNewPictures, ...formerPictures],
                    },
                }
```

In my case (and in the case provided by API Platform doc), I wonder how I could achieve sending directly FormData body instead of stringified JSON.

I begin to write something like this, but I can't find a way to not recreating my own convertReactAdminRequestToHydraRequest and convertHydraResponseToReactAdminResponse : 
```
export const addUploadFeature = requestHandler => (type, resource, params) => {
  if ( (type === UPDATE || type === CREATE) && resource === 'media_objects' && typeof params.data.contentUrl != 'undefined' && typeof params.data.contentUrl.rawFile != 'undefined' ) {
    const encodeData = (data) => {
      const body = new FormData();
      body.append('file', data.contentUrl.rawFile);
      body.append('contentUrl', data.contentUrl.src);
      body.append('name', data.name);
      return body;
    };
    
    // it's this part that needs to bypass
    return convertReactAdminRequestToHydraRequest(type, resource, params)
      .then(({url, options}) => fetchHydra(url, options))
      .then(response =>
        convertHydraResponseToReactAdminResponse(type, resource, response),
      );
  }
  // for other request types and resources, fall back to the default request handler
  return requestHandler(type, resource, params);
};

// baseDataProvider is the one provided by @api-platform/admin
// apiDocumentationParser is just a decorated version for file input and custom routes, like explained in the docs
export default (api, apiDocumentationParser) => addUploadFeature(baseDataProvider(api.entrypoint, fetchHydra, apiDocumentationParser));
```

Finally, I don't know if rewriting a lot of code for just bypassing JSON.stringify is not worse than providing a simple way to achieve this.

I can help to provide example in the right way for updating API Platform docs, but I need you to point me the right direction 😃 

Thanks for your help

src/hydra/dataProvider.js
rkcreation: I agree for the code, I'd just restored the original one.

For the test, I try to find the right solution first

mrkonovalov:
I changed the URL API `/images` to `/images/upload`. Added method `create` in dataProvider for images with the transformation of Object to FormData. This life hack worked because `resource` in HydraDaraProvider wasn't found here: https://github.com/api-platform/admin/blob/v2.2.1/src/hydra/dataProvider.js#L312

alanpoulain:
See https://api-platform.com/docs/admin/file-upload/ for the new way to handle file upload.

PR #257, opened 2019-11-12
feat: first steps to migrate to v3 or react-admin (still in beta)

Gregcop1:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Migration to React admin (v.3)

PR #263, opened 2020-01-03
React-admin 3

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Supersede https://github.com/api-platform/admin/pull/257.

Upgrade the Admin for react-admin 3.

It will be a new 2.0 version because of the BC.

Things done:
- Use hooks instead of classes.
- `AdminGuesser` can be used instead of `HydraAdmin` (to make it clear it's not related to Hydra).
- A custom `resourceSchemaAnalyzer` can be pass as a prop of `AdminGuesser`.
- A custom `history` can be pass as a prop.
- The Hydra data provider is using the new type.
- Unused code removed and clean up.


alanpoulain:
@fzaninotto It would be great if you could have a look :slightly_smiling_face:

src/hydra/HydraAdmin.js
```diff
@@ -1,50 +1,165 @@
...
+  return <AdminUI {...rest}>{children}</AdminUI>;
+};
 
 const HydraAdmin = ({
   entrypoint,
   dataProvider = dataProviderFactory(entrypoint),
-  i18nProvider,
   authProvider,
+  i18nProvider,
+  history = createHashHistory(),
```
mleczakm: WDYT about making it configurable to support also browser history?

alanpoulain: It's a prop so you can pass whatever you want (for example an instance of `createBrowserHistory`).

yudhir:
Thanks, looking Forward .

src/hydra/HydraAdmin.js
mleczakm: You should add info about this change to "Things done:" because it was not possible before :)

alanpoulain: Sure :slightly_smiling_face:

Deuchnord:
Tested on a fresh project after an `npm uninstall @api-platform/admin && npm install --save alanpoulain/admin#ra3 && docker-compose up --build`, didn't found any regression. :+1:

src/AdminGuesser.js
```diff
@@ -16,51 +21,66 @@ const displayOverrideCode = resources => {
...
+  includeDeprecated = false,
 
-AdminGuesser.defaultProps = {
-  includeDeprecated: false,
-  theme: createMuiTheme({
+  appLayout,
+  layout = Layout,
+  loginPage,
+  locale,
+  theme = createMuiTheme({
```
fzaninotto: prefer `defaultProps` to default values in the component props signature, because `defaultProps` are only created once, whereas default values are created each time the component renders. You don't know how many times you component may render, and returning a new value for the `theme` prop will rerender the whole tree even if the two themes are shallow equal.

src/AdminGuesser.js
```diff
@@ -71,11 +91,76 @@ AdminGuesser.defaultProps = {
...
+  const [resources, setResources] = useState();
+  const [fetching, setFetching] = useState(true);
+  const [error, setError] = useState();
+
+  if (appLayout) {
+    console.warn(
+      'You are using deprecated prop "appLayout", it was replaced by "layout", see https://github.com/marmelab/react-admin/issues/2918',
+    );
+  }
+  if (loginPage === true && process.env.NODE_ENV !== 'production') {
```
fzaninotto: why warn only in production here and not for the other two warnings?

src/AdminGuesser.js
```diff
@@ -71,11 +91,76 @@ AdminGuesser.defaultProps = {
...
+    );
+  }
+  if (locale) {
+    console.warn(
+      'You are using deprecated prop "locale". You must now pass the initial locale to your i18nProvider',
+    );
+  }
+
+  useEffect(() => {
+    dataProvider
```
fzaninotto: I suggest you add a detection for the `dataProvider.introspect()` method existence here, and throw a nice Exception in case a developer uses a non-hydra dataProvider. As this method is not in the standard `dataProvider` signature, you need to help developers a bit.

src/AdminGuesser.test.js
```diff
@@ -29,7 +26,7 @@ describe('<AdminGuesser />', () => {
   test('renders loading', () => {
     const tree = renderer.render(
       <Provider store={store}>
-        {AdminGuesserComponent({}, {loading: true})}
+        {AdminGuesserComponent({fetching: true})}
```
fzaninotto: same, and same for the other tests

src/hydra/dataProvider.js
```diff
@@ -360,38 +359,44 @@ export default (
...
-        return convertReactAdminRequestToHydraRequest(type, resource, params)
-          .then(({url, options}) => httpClient(url, options))
-          .then(response =>
-            convertHydraResponseToReactAdminResponse(type, resource, response),
-          );
-    }
+    getMany: (resource, params) =>
+      Promise.all(
+        params.ids.map(id =>
+          reactAdminDocumentsCache[id]
```
fzaninotto: I'm not sure what that is, but react-admin already uses an internal cache on the client-side. If the dataProvider adds another layer of caching, this can cause stale data.

Also, all other methods call `fetchAPI` directly, why should this one use a cache and not the others?

src/hydra/dataProvider.test.js
```diff
@@ -78,36 +77,38 @@ describe('Transform a React Admin request to an Hydra request', () => {
...
-      expect(searchParams[5]).toEqual([
-        'nested_array.nested[0]',
-        '/nested_iri/1',
-      ]);
-      expect(searchParams[6]).toEqual([
-        'nested_array.nested[1]',
-        '/nested_iri/2',
-      ]);
-    });
+    return dataProvider
```
fzaninotto: I prefer using the `async/await` syntax to avoid deeper nesting, but that's just a preference.

src/hydra/resourceSchemaAnalyzer.js
```diff
@@ -0,0 +1,68 @@
+const getReferenceNameField = reference => {
```
fzaninotto: This entire file is hard to read and would greatly benefit from input/output examples in jsDoc.

src/Introspecter.js
```diff
@@ -0,0 +1,104 @@
+import React, {useContext, useEffect, useState} from 'react';
+import PropTypes from 'prop-types';
+import {Loading, useDataProvider} from 'react-admin';
+import ResourceSchemaAnalyzerContext from './ResourceSchemaAnalyzerContext';
+
+const IntrospecterComponent = ({
```
fzaninotto: bad name for a component. By the way, why does this file split its logic in two components that are entirely intricated?

src/Introspecter.js
```diff
@@ -0,0 +1,104 @@
...
+}) => {
+  const dataProvider = useDataProvider();
+  const resourceSchemaAnalyzer = useContext(ResourceSchemaAnalyzerContext);
+  const [resources, setResources] = useState();
+  const [fetching, setFetching] = useState(true);
+  const [error, setError] = useState();
+
+  useEffect(() => {
+    dataProvider
+      .introspect()
```
fzaninotto: Am I understanding correctly: as `Introspecter` is called in every `Guesser`, this will fetch the entire hydra schema once for each resource, field, input? If so, that's a huge waste of bandwidth and a major slowdown. Especially since you've already fetched the resources in AdminGuesser.

src/ListGuesser.js
```diff
@@ -24,59 +25,62 @@ const displayOverrideCode = (resourceSchema, fields) => {
...
-  }
+  useEffect(() => {
+    if (resourceSchema) {
+      const resolvedOrderParameters = resourceSchemaAnalyzer.getOrderParametersFromResourceSchema(
+        resourceSchema,
+      );
 
-  render() {
-    const {resourceSchema, fields, ...props} = this.props;
+      setOrderParameters(resolvedOrderParameters);
```
fzaninotto: I don't understand why you setState here even though `resolvedOrderParameters` might be empty, and you'll have to call the `resourceSchema` just next line to set it again.

src/ListGuesser.js
```diff
@@ -24,59 +25,62 @@ const displayOverrideCode = (resourceSchema, fields) => {
...
-      <List {...props}>
-        <Datagrid>
-          {props.children}
-          {props.hasShow && <ShowButton />}
-          {props.hasEdit && <EditButton />}
-        </Datagrid>
-      </List>
-    );
+  if (!children) {
+    children = fields.map(field => (
```
fzaninotto: don't set a new value for a function argument

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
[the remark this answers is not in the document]
Wirone: @fzaninotto The whole purpose of this file is to define common options for files' formatting, so everyone working on the project can (with support from IDEs) enforce the same configuration without any manual work. With this you won't get unintentional whitespace diffs. It should be commited to work.

The question is: do project maintainers want this file and current configuration fits project's standards?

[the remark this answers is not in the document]
dunglas: It looks good to me (we have it on every other API Platform projects).

src/AdminGuesser.js
```diff
@@ -1,13 +1,18 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
```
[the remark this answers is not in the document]
dunglas: This project already uses Prettier: https://github.com/api-platform/admin/blob/master/package.json#L50

Running `yarn fix` should do the trick.

src/Introspecter.js
dunglas: For testing purpose, IIRC.

src/Introspecter.js
dunglas: The introspection logic is only called once then fetched from a local cache: https://github.com/api-platform/admin/blob/b022e157db00378df57cf22e544bf33ba55dd173/src/hydra/dataProvider.js#L116

src/AdminGuesser.test.js
```diff
@@ -12,14 +12,11 @@ describe('<AdminGuesser />', () => {
...
       <Provider store={store}>
-        {AdminGuesserComponent(
-          {},
-          {
-            error: 'Failed to fetch documentation',
-            data: {resources},
-            loading: false,
-          },
-        )}
+        {AdminGuesserComponent({
```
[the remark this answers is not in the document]
alanpoulain: I think it has been done this way because the `ShallowRenderer` from `react-test-renderer` only allows to render the component "one level deep" and using the `Provider` is necessary.

src/AdminGuesser.js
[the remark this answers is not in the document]
alanpoulain: The rule `bracketSpacing: false` had to be removed to fix it.

src/AdminGuesser.js
```diff
@@ -1,13 +1,18 @@
...
 import PropTypes from 'prop-types';
-import {Admin, Loading, Query, TranslationProvider} from 'react-admin';
-import {createMuiTheme} from '@material-ui/core/styles';
-import {Layout} from './layout';
+import {AdminContext, AdminUI, Error as ErrorUI, Loading} from 'react-admin';
```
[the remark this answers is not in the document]
alanpoulain: It's because I use `new Error`.

src/AdminGuesser.js
```diff
@@ -16,51 +21,66 @@ const displayOverrideCode = resources => {
...
-  if (loading) {
-    return (
-      <TranslationProvider>
-        <Loading />
-      </TranslationProvider>
-    );
+export const AdminGuesserComponent = ({
+  children,
+  includeDeprecated,
+
```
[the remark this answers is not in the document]
alanpoulain: It was to organize the props:
- for the `AdminContext`,
- for the `AdminResourcesGuesser`,
- for the `AdminUI`,
- for the `SchemaAnalyzerContext`.

But maybe it's a bad idea?

src/AdminGuesser.js
alanpoulain: It seems `defaultProps` will be deprecated for function component: https://github.com/reactjs/rfcs/pull/107.
That's why I chose to do it like this.
Maybe I can do the instanciation before though for performance reasons.

src/AdminGuesser.js
alanpoulain: Makes sense but FYI it's how it's done in react-admin: https://github.com/marmelab/react-admin/blob/56b1d4096a5e68e442d29b1b7b12699d3bfc3f4b/packages/react-admin/src/Admin.tsx#L107-L121.

src/hydra/dataProvider.js
alanpoulain: It has been done like this since a long time. Maybe @dunglas could answer this better than me?

src/ListGuesser.js
alanpoulain: It seems it's because sometimes the order parameters are already in the parsed schema, and sometimes it's not and a supplementary call to the API is needed in order to retrieve them. I'm not sure why but maybe @dunglas or @mauchede can answer?

alanpoulain:
The `Introspecter` now uses the `useQueryWithStore` hook in order to work under optimistic mode.

src/ListGuesser.js
dunglas: Actually, Hydra doesn't support (yet) to expose available filters in `docs.jsonld`, this information is only available in the metadata of the collection, so a call to the 1st page is needed to retrieve these parameters.

src/AdminGuesser.test.js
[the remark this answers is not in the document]
fzaninotto: You mean in tests? Then don't use shallow rendering, use react-testing-library instead.

src/AdminGuesser.js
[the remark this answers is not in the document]
fzaninotto: Then use a comment instead of a line break to explain your organization

src/AdminGuesser.js
fzaninotto: yes, you'll need to move the createMuiTheme outside

src/AdminGuesser.js
fzaninotto: Then it's worth an issue in the react-admin repo!

src/AdminGuesser.test.js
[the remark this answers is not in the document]
alanpoulain: The `Provider` was not needed anymore so we can keep the `react-test-renderer`.

src/AdminGuesser.js
alanpoulain: Or a PR :slightly_smiling_face: https://github.com/marmelab/react-admin/pull/4279

PR #269, opened 2020-01-19
Use array notation for exists, range and date filters

### LINKED ISSUE #260: Range filter changes [] style to dot notation
If use standard filters to achieve range filter functionality, i.e.:
```
<Filter {...props}>
    <SelectInput
        source='year[gte]'
        label="Year from"
        choices={getYears()}
        alwaysOn
    />
</Filter>
```
as a result you will get api request: year.gte=2002 instead of required year[gte]=2002

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #260
| License       | MIT
| Doc PR        | N/A

React Final Form doesn't make the distinction between array notation and dot notation (see https://final-form.org/docs/final-form/field-names).

Since some filters in API Platform use array notation (exists, range and date filters), the data provider needs to check if some specific keywords for these filters are used.

Hence using a hardcoded list of keywords like this.

The drawbacks of this method is the need to maintain this list and the risk of conflict with property names.

A second solution would be to use a special syntax like curly brackets (like `date{after}`) but I don't think it's an user-friendly solution since it needs to be documented.

A third solution would be to use something like `HydraFilter` which would replace array notation to curly notation (and it would use the second solution too).

Maybe we can introduce the last solutions for special cases like if the exists filter parameter name is modified. WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.0.1
 
 * Use readable and writable fields instead of fields
+* Use array notation for exists, range and date filters
 
 ## 2.0.0
 
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -235,6 +235,26 @@ export default (
             }
 
             Object.keys(filterValue).forEach(subKey => {
+              if (
+                rootKey === 'exists' ||
+                [
+                  'after',
+                  'before',
+                  'strictly_after',
+                  'strictly_before',
+                  'lt',
+                  'gt',
+                  'lte',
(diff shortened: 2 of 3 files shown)


soyuka:
Using `[]` is right imo. I'm okay for this specific solution :+1:.

mleczakm:
When will it be merged, if there are no objections? We have already same fix in our custom-filter and it works as expected, can't wait to have this fix on prod since filtering over dates is not working.

PR #280, opened 2020-02-28
Total items undefined

### LINKED ISSUE #276: Error after enabling Partial Pagination
I have a table with 7 million rows, so pagination is very slow since it is running the count query. So I have set the partial pagination to true in my Global configuration.

But in the admin section when I select this table I am getting the below error.

Given action "RA/CRUD_GET_LIST_SUCCESS", reducer "total" returned undefined. To ignore an action, you must explicitly return the previous state. If you want this reducer to hold no value, you can return null instead of undefined.

Any one has any idea, any help will be much appreciated. Thanks.

Edit 1: More info and Steps to Reproduce added.

1. A normal request would return a Response with "totalItems"
`"hydra:totalItems": 3328`
2. When you enable Partial Pagination then we don't have "totalItems" in the Response anymore. I Think this is a problem in the admin since it is always expecting "totalItems".

**Steps to Reproduce:**
1. Enable Partial Pagination
`# api/config/packages/api_platform.yaml
api_platform:
    collection:
        pagination:
            partial: true # Disabled by default`
2. Browse the admin and you will get the error.

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #276 
| License       | MIT
| Doc PR        | na

I don't think that this fix is enough as it will disable pagination client side

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -358,7 +358,7 @@ export default (
               ),
             ),
           )
-          .then(data => ({ data, total: response.json['hydra:totalItems'] }));
+          .then(data => ({ data, total: response.json['hydra:totalItems'] === undefined ? null : response.json['hydra:totalItems']}));
 
       case DELETE:
         return Promise.resolve({ data: { id: null } });


alanpoulain:
Supersed by https://github.com/api-platform/admin/pull/280.

PR #292, opened 2020-04-13
Removing trace from body when Error happend

### LINKED ISSUE #291: `trace` key is causing jsonld.expand to fetch wrong urls
**API Platform version(s) affected**: 2.1.0

**Description**  
When endpoint returns an error with a trace (DEBUG)
image
image

`getDocumentationUrlFromHeaders` is failing because it's trying to fetch `/api/array` it happens because `_retrieveContextUrls` is visiting all keys and somehow finds this weird endpoint.

oleg-andreyev:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #291
| License       | MIT

### DIFF
--- a/src/hydra/fetchHydra.js
+++ b/src/hydra/fetchHydra.js
@@ -30,8 +30,11 @@ export default (url, options = {}) => {
     const status = data.response.status;
 
     if (status < 200 || status >= 300) {
+      const body = data.body;
+      delete body.trace;
+
       return promises
-        .expand(data.body, {
+        .expand(body, {
           base: getDocumentationUrlFromHeaders(data.response.headers),
         })
         .then((json) => {


src/hydra/fetchHydra.js
```diff
@@ -30,8 +30,11 @@ export default (url, options = {}) => {
     const status = data.response.status;
 
     if (status < 200 || status >= 300) {
+      const body = data.body;
+      delete body.trace;
```
soyuka, line 34: shouldn't we log this though?

oleg-andreyev: It will be logged by backend + you'll be able to see it in Network Tab, but for Production Mode there won't be any trace

soyuka:
thanks @oleg-andreyev !

Issue #144, opened 2019-01-11
Support for Mercure


dunglas:
The client generator supports Mercure, but it's not yet the case of the admin. It would be a nice addition!

dunglas:
Should use https://github.com/marmelab/react-admin/tree/master/packages/ra-realtime

L-Luciano:
Hello,

Any news for this enchancement ?

Regards,

andrekutianski:
This will be nice!

dunglas:
Hi! We haven't implemented this yet but the Marmelab team provide this feature as part of the (paid) React Admin Enterprise package: https://marmelab.com/ra-enterprise/modules/ra-realtime

alanpoulain:
Will be supported in 2.7.0.

PR #297, opened 2020-05-11
Add partial pagination support

### LINKED ISSUE #276: Error after enabling Partial Pagination

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #276
| License       | MIT
| Doc PR        | N/A

Add a partial pagination support by adding a custom previous / next pagination component when needed.

It can be considered as a bug fix and as a new feature.

It uses some kind of hack for the hydra data provider to transmit pagination information to the pagination component, by using the `total` prop.

I don't think there is a better way to do so without modifying react-admin.

@fzaninotto / @djhi maybe it would be interesting if you could give your point of view?

### DIFF
--- a/src/ListGuesser.js
+++ b/src/ListGuesser.js
@@ -4,6 +4,7 @@ import { Datagrid, List, EditButton, ShowButton } from 'react-admin';
 import FieldGuesser from './FieldGuesser';
 import FilterGuesser from './FilterGuesser';
 import Introspecter from './Introspecter';
+import Pagination from './list/Pagination';
 
 const displayOverrideCode = (schema, fields) => {
   if (process.env.NODE_ENV === 'production') return;
@@ -58,7 +59,7 @@ export const IntrospectedListGuesser = ({
   }
 
   return (
-    <List {...props}>
+    <List pagination={<Pagination />} {...props}>
       <Datagrid>
         {fieldChildren}
         {props.hasShow && <ShowButton />}
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -360,7 +360,16 @@ export default (
               ),
             ),
           )
-          .then((data) => ({ data, total: response.json['hydra:totalItems'] }));
+          .then((data) => ({
+            data,
+            total:
+              response.json?.['hydra:totalItems'] ||
+              (response.json?.['hydra:view']
+                ? response.json['hydra:view']?.['hydra:next']
+                  ? -2 // there is a next page
+                  : -1 // no next page
+                : -3), // no information
+          }));
 
       case DELETE:
         return Promise.resolve({ data: { id: null } });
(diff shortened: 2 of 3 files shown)


alanpoulain:
Merging for now but you can comment on this PR :slightly_smiling_face:

fzaninotto:
I'm not sure I understand the problem. Can you please elaborate?

alanpoulain:
Sure. In API Platform we have a partial pagination support: https://api-platform.com/docs/core/pagination/#partial-pagination. It means the API doesn't tell how many pages there are, but only if there is a previous / next page.
When activated for a resource, the previous behavior was buggy because the react-admin pagination component was used and it needs to know the number of pages to work correctly.

To fix it, I created a pagination wrapper. Its behavior changes according to the information on the pagination: if the total is given, its the classical pagination, if not it's the previous / next pagination. However, in order to know if the next button needs to be displayed or not, the hydra data provider needs to tell the component if there is a next page or not. In the response format of the `getList` (`{ data: {Record[]}, total: {int}, validUntil?: {Date} }`), there is no way to add more information (at least I don't see one) than the data and the total.

So I needed to use some tricks (use the `total`), to send the information about the next page. Maybe a way would be to add an optional parameter (`extraInformation` for instance?) in the response format and to pass it as a prop to the pagination component?

fzaninotto:
Thanks, it's much clearer. 

We would need an additional field in the response (I'd call it `meta`) to carry that kind of information. But it requires updating quite a lot of things (TypeScript types, dataProvider signatures, dataProvider hooks, controllers, views, pagination) and I find your hacky solution not too hacky ;)

So let's say that we won't address that use case in react-admin for now.

alanpoulain:
OK. And if provide a PR in react-admin, would it be alright? Because I think it would be useful for the Mercure support too (https://github.com/api-platform/admin/issues/144). For Mercure, a link to the hub is sent in a response header. But the data provider would have no way to send this link to the components needed to subscribe to the updates. And I think we can generalize this to GraphQL subscriptions too.

fzaninotto:
Sure: a good way to see if it's not too big a change is to give it a try. Make sure to PR against the `next` branch though.

PR #299, opened 2020-05-16
Add embedded support

### LINKED ISSUE #279: Edit a record with a complex structure is not working
**API Platform version(s) affected**: v2.5.4 admin - v2.0.1

**Description**  
According to React Admin's documentation:
**Tip:** If you edit a record with a complex structure, you can use a path as the source parameter. For instance, if the API returns the following ‘book’ record:
```
{
    id: 1234,
    title: 'War and Peace',
    author: {
        firstName: 'Leo',
        lastName: 'Tolstoi'
    }
}
```
Then you can display a text input to edit the author first name as follows:
`<TextInput source="author.firstName" />`

As Api Platform admin's dataProvider replaces embedded objects with their IRI. Value from the record is not set to the input in Edit component.

**Save itself for such a field is working, but data preload - not**

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (slightly and only if embeddeds are used)
| Deprecations? | no
| Tickets       | fixes #279
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1112

Alternative of https://github.com/api-platform/admin/pull/107.

Since `@api-platform/api-doc-parser` can now tell if a field is an embedded one, the admin can be slightly smarter and not using a reference for it.

If it's an array of embeddeds, the previous behavior stays the same: objects are replaced by their IRI and references are used in `FieldGuesser` and `InputGuesser`.

However, if it's an embedded field, the behavior changes depending of the value of `useEmbedded` in the Hydra data provider:
- if `false`, the IRI of the embedded will be displayed as a text (**BC**: it was a reference before, you can get back this behaviour by replacing the guessers by references) in `FieldGuesser` and `InputGuesser`.
- if `true`, the embedded object will be displayed as a text in `FieldGuesser` and `InputGuesser`.

The `useEmbedded` is `false` by default to not cause too much breaking changes, but it will not exist anymore in 3.0 (and the behavior will be like it was always `true`).

It means that, if you enable `useEmbedded` in the Hydra data provider, you can now use the **dot notation of React Admin for complex structures**!

However, the admin cannot determine the structure of the embedded and the fields in it. It's because the Hydra documentation doesn't give this information. In order to display them, you need to use directly the React Admin components.

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -41,6 +41,7 @@ const reactAdminDocumentsCache = new Map();
  * @param {Object} document
  * @param {bool} clone
  * @param {bool} addToCache
+ * @param {bool} useEmbedded
  *
  * @return {ReactAdminDocument}
  */
@@ -48,6 +49,7 @@ export const transformJsonLdDocumentToReactAdminDocument = (
   document,
   clone = true,
   addToCache = true,
+  useEmbedded = false,
 ) => {
   if (clone) {
     // deep clone documents
@@ -72,7 +74,7 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           false,
         );
       }
-      document[key] = document[key]['@id'];
+      document[key] = useEmbedded ? document[key] : document[key]['@id'];
 
       return;
     }
@@ -116,6 +118,7 @@ export default (
   entrypoint,
   httpClient = fetchHydra,
   apiDocumentationParser = parseHydraDocumentation,
+  useEmbedded = false, // remove this parameter for 3.0 (as true)
 ) => {
   /** @type {Api} */
   let apiSchema;
@@ -349,8 +352,13 @@ export default (
       case GET_MANY_REFERENCE:
         // TODO: support other prefixes than "hydra:"
         return Promise.resolve(
-          response.json['hydra:member'].map(
-            transformJsonLdDocumentToReactAdminDocument,
+          response.json['hydra:member'].map((document) =>
+            transformJsonLdDocumentToReactAdminDocument(
+              document,
+              true,
+              true,
+              useEmbedded,
+            ),
           ),
         )
           .then((data) =>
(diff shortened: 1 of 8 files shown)


src/hydra/dataProvider.js
```diff
@@ -72,7 +74,7 @@ export const transformJsonLdDocumentToReactAdminDocument = (
           false,
         );
       }
-      document[key] = document[key]['@id'];
+      document[key] = useEmbedded ? document[key] : document[key]['@id'];
```
Darkweak, line 77:
```suggestion
      document[key.toString()] = useEmbedded ? document[key.toString()] : document[key.toString()]['@id'];
```
To avoid Object Injection Sink IMO

alanpoulain: It cannot be done like this because React Admin data providers need to give record objects: https://marmelab.com/react-admin/DataProviders.html#response-format

Darkweak: It doesn't affect return, your key object might be a string right ?

alanpoulain: Oh yes, you're right sorry. But `Object.keys` returns only strings so I don't think it's necessary.

yannick-loeffler:
Hey guys,

thanks for your work! Unfortunately, those changes doesn't seem to work for me (or I am doint something totally wrong). I am using "useEmbedded=true" and my embedded objects are not resolved :(

I also followed your docu in 
https://api-platform.com/docs/admin/handling-relations/#display-a-field-of-an-embedded-relation

Further details you can finde here:
https://stackoverflow.com/questions/61704585/access-property-of-a-nested-record

alanpoulain:
@loeyan the embedded support was not done for the array of embeddeds yet. The PR to add it is here: https://github.com/api-platform/admin/pull/302.

PR #306, opened 2020-06-08
Fix issue #305

### LINKED ISSUE #305: Show and edit use incorrect api endpoint
**API Platform version(s) affected**: 2.1.1

**Description**  
Going to the show or edit page of an entry does not use the correct API endpoint. Saving an entry uses the same URL and has therefore the same incorrect behavior.

**How to reproduce**  
Core application is hosted at localhost
API is hosted at localhost/api
Admin is hosted at localhost/admin

Entity type example is "members". The list view displays correctly and fetches data from localhost/api/members. Show and edit view should fetch data from localhost/api/members/id, but these calls go to localhost/id instead.

Entrypoint for HydraAdmin is configured to be localhost/api.

**Possible Solution**  
Change buildup of the itemURL in the hydraDataProvider. I was able to fix the issue locally. I am now working on a pull request.

wim-beck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #305
| License       | MIT
| Doc PR        | /

No tests were added. Existing tests passed.

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -188,7 +188,7 @@ export default (
   const convertReactAdminRequestToHydraRequest = (type, resource, params) => {
     const entrypointUrl = new URL(entrypoint, window.location.href);
     const collectionUrl = new URL(`${entrypoint}/${resource}`, entrypointUrl);
-    const itemUrl = new URL(params.id, entrypointUrl);
+    const itemUrl = new URL(`${entrypoint}/${resource}/${params.id}`, entrypointUrl);
 
     switch (type) {
       case CREATE:
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -188,7 +188,10 @@ export default (
   const convertReactAdminRequestToHydraRequest = (type, resource, params) => {
     const entrypointUrl = new URL(entrypoint, window.location.href);
     const collectionUrl = new URL(`${entrypoint}/${resource}`, entrypointUrl);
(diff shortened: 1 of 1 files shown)


alanpoulain:
The admin is working with IRIs, not IDs: `params.id` should already contain the resource.

wim-beck:
Thanks for the info.
My api endpoint is not returning IRIs because I implemented a custom doctrine ORM extension to filter out some fields if a non-admin user performs a GET request. If I disable the extension, the IRI is returned (along with all other fields). If I enable the extension, the subset of properties is returned, but the IRI is not included, just the id.

I tried adding "@ApiProperty(identifier=true)" to my ApiResource class, but this does not help either.

How can I fix this?

alanpoulain:
Probably because the Hydra / JSON-LD normalizer is not called correctly. You can ask for help in the Slack channel or in Stack Overflow (https://api-platform.com/community).

wim-beck:
Thanks for the pointer.

My solution for people having the same issue:

The culprit was the select clause in my custom ORM extension. I had a select clause that took the id, and some basic properties when a normal user requested the info and an addSelect to select additional properties for an admin user. The id was returned, but for some reason not recognized as the id of the entity and not normalized to IRI format.

After removing the select clauses from my query builder, the proper IRI was returned. 

I then added security to filter out fields based on the user requesting the info by implementing a context builder (https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-dynamically).

Issue #281, opened 2020-03-05
Next.js - Invariant failed: Hash history needs a DOM (SSR)


Vadorequest:
**API Platform version(s) affected**: 2.0.1

**Description**  
When running the app using Next.js (isomorphic framework), it crashes with `Invariant failed: Hash history needs a DOM`.

**How to reproduce**  
Source code (OSS): https://github.com/UnlyEd/next-right-now-admin/pull/1

**Possible Solution**  
I thought that was related to SSR and tried to render `HydraAdmin` only on the client but it didn't work.

Note that using react-admin directly works fine:
Source code: https://github.com/UnlyEd/next-right-now-admin/blob/nrn-admin-poc/src/pages/index.tsx
Demo: https://next-right-now-admin.now.sh/#/Product

**Additional Context**  
```
Invariant failed: Hash history needs a DOM
Error: Invariant failed: Hash history needs a DOM
    at invariant (/Users/vadorequest/dev/next-right-now-admin/node_modules/tiny-invariant/dist/tiny-invariant.cjs.js:13:11)
    at createHashHistory (/Users/vadorequest/dev/next-right-now-admin/node_modules/history/cjs/history.js:558:16)
    at Object.<anonymous> (/Users/vadorequest/dev/next-right-now-admin/node_modules/@api-platform/admin/lib/AdminGuesser.js:77:55)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Module.require (internal/modules/cjs/loader.js:1040:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/Users/vadorequest/dev/next-right-now-admin/node_modules/@api-platform/admin/lib/index.js:91:45)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Module.require (internal/modules/cjs/loader.js:1040:19)
    at require (internal/modules/cjs/helpers.js:72:18)
```

alanpoulain:
[...]

PR #309, opened 2020-06-22
fix: Define history only in a browser context

soyuka:
See also #281

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

### DIFF
--- a/src/AdminGuesser.js
+++ b/src/AdminGuesser.js
@@ -60,7 +60,6 @@ export const AdminResourcesGuesser = ({
   return <AdminUI {...rest}>{resourceChildren}</AdminUI>;
 };
 
-const defaultHistory = createHashHistory();
 const defaultTheme = createMuiTheme({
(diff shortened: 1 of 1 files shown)


pawelsmolenski:
@soyuka Please, try to readjust your code, so checks of this MR will pass... As a motivation I may say that this fix will also fix a problem when history strategy other than "hash" is used. Now, including (indirectly) an `AdminResourcesGuesser` component causes that a hash is always added to the URL on application start in a browser. As a result, this behavior messes the browser history what may and causes a waterfall of other problems – especially on the login page.

In other words: removing obligatory initialization of Hash History by not calling createHashHistory() when another history service is provided will stop HydraAdmin from adding hash to the end of URL on startup.

PR #317, opened 2020-07-28
fixing data-provider

### LINKED ISSUE #316: Error message is lost after 2.2.1 release
**API Platform version(s) affected**: 2.2.1 

**Description**  
When executing `apiDocumentationParser` before 2.2.1, I would get error message with explanation why exactly is wrong. 

**How to reproduce**  
`apiDocumentationParser` should throw an Error

**Error 2.2.1**
```
Error: Cannot fetch API documentation:
Have you verified that CORS is correctly configured in your API?
```

**Before 2.2.1**
```
Error: Cannot fetch API documentation:
Unable to find the URL for "https://127.0.0.1:8000/api/docs.jsonld#Entrypoint/transaction", make sure your api resource has at least one GET item operation declared.
Have you verified that CORS is correctly configured in your API?
```

oleg-andreyev:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #316
| License       | MIT

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -467,7 +467,7 @@ export default (
               apiSchema = api;
               return { data: api, customRoutes };
             })
-            .catch(({ error, status }) => {
+            .catch((error, status) => {
               throw new Error(
                 'Cannot fetch API documentation:\n' +
                   (error


src/hydra/dataProvider.js
```diff
@@ -467,7 +467,7 @@ export default (
               apiSchema = api;
               return { data: api, customRoutes };
             })
-            .catch(({ error, status }) => {
+            .catch((error, status) => {
```
oleg-andreyev, line 470: This is because `apiDocumentationParser` does not throw an object with error and status fields.

alanpoulain:
Did you check if it works when there is a CORS issue?

src/hydra/dataProvider.js
soyuka: I've never seen a catch with 2 arguments, imo it should be something like:

```
catch((error) => {
  const status = error.status
  // then check later if needed
})
```

oleg-andreyev: that's how it was before 2.2.1, but somehow `status` was added. 
https://github.com/api-platform/admin/compare/v2.2.0...v2.2.1#diff-00342c75206eee49e87735785d9c2e74L470

soyuka:
see #318

PR #318, opened 2020-08-05
fix: catch error with status

soyuka:
see https://github.com/api-platform/admin/pull/317/files

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -467,11 +467,12 @@ export default (
               apiSchema = api;
               return { data: api, customRoutes };
             })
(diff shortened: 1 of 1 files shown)


src/hydra/dataProvider.js
```diff
@@ -467,7 +467,8 @@ export default (
               apiSchema = api;
               return { data: api, customRoutes };
             })
-            .catch(({ error, status }) => {
+            .catch((error) => {
+              const { status } = error;
```
alanpoulain, line 471: Deconstruct message too?

alanpoulain:
Could you test it when CORS are not enabled?

PR #323, opened 2020-09-14
WIP: feat(hydra): added custom param for data provider

justinezahiri:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This contribution aims to improve the dataProvider, to add custom params in url by customizing it.
For example adding `pagination=true`.

PR #328, opened 2020-10-27
Data provider update many params

### LINKED ISSUE #327: useUpdateMany hook making empty put call
**API Platform version(s) affected**: "@api-platform/admin": "^2.2.1",

**Description**  
I'm having a problem using useUpdateMany's hook for React Admin.

If I try to use useUpdate it works but with many, it's sending an empty information, here goes some code:

```
    const BulkActions = props => {
    
        const refresh = useRefresh();
        const notify = useNotify();
    
        const unselectAll = useUnselectAll();
        const [updateMany, { loading, error }] = useUpdateMany(
            props.resource,
            props.selectedIds,
            { totalCommission: '100' },
            {
                onSuccess: () => {
                    refresh();
                    notify('Information updated', 'info', {}, true);
                    unselectAll(props.resource);
                },
                onFailure: error => notify('Error: posts not updated', 'warning'),
                undoable: true
            }
        );
    
    
        return (
            <Fragment>
                <Button label="Confirm" variant="contained" onClick={ updateMany } disabled={loading} />
            </Fragment>
        );
    };
```

This is the dataProvider

   ```
 import React from "react";
    import { Redirect, Route } from "react-router-dom";
    import {API_BASE_URL} from '../constants/constants'
    import { hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydra } from "@api-platform/admin";
    import { parseHydraDocumentation } from "@api-platform/api-doc-parser";
    
    const entrypoint = API_BASE_URL;
    
    const fetchHeaders = { Authorization: `Bearer ${localStorage.getItem("token")}` };
    const fetchHydra = (url, options = {}) => baseFetchHydra(url, {
        ...options,
        headers: new Headers(fetchHeaders),
    });
    
    const apiDocumentationParser = entrypoint => parseHydraDocumentation(entrypoint, { headers: new Headers(fetchHeaders) })
        .then(
            ({ api }) => ({ api }),
            (result) => {
                const {api, status} = result;
                if (result.status === 401) {
                    return Promise.resolve({
                        api,
                        status,
                        customRoutes: [
                            <Route path="/" render={() => <Redirect to="/login" />} />,
                        ],
                    });
                }
    
                return Promise.reject(result);
            },
        );
    let dataProvider = baseHydraDataProvider(entrypoint, fetchHydra, apiDocumentationParser, true);
    
    export default dataProvider;

```

**How to reproduce**  
 I tried to update a bulk action with some data

**Possible Solution**  
Inside the dataProvider.js the return updateMany is fetching id without data
```
updateMany: (resource, params) => Promise.all(params.ids.map(id => fetchApi(_reactAdmin.UPDATE, resource, {
      id
    }))).then(() => ({
      data: []
    }))
```
Data need to be added there

### PR DESCRIPTION
**Pull request by braianj**
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #327
| License       | MIT

Fix issue when use the hook useUpdateMany sending parameters.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@api-platform/admin",
-  "version": "2.2.3",
+  "version": "2.2.4",
   "description": "Automatic administration interface for Hydra-enabled APIs.",
   "files": [
     "*.md",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@api-platform/admin",
-  "version": "2.2.4",
+  "version": "2.2.3",
   "description": "Automatic administration interface for Hydra-enabled APIs.",
   "files": [
     "*.md",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.4
+
+* Fix updateMany to send parameters into data fetch
+
 ## 2.2.3
 
 * Hydra: display empty page when `hydra:totalItems` is 0
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -451,7 +451,7 @@ export default (
     update: (resource, params) => fetchApi(UPDATE, resource, params),
     updateMany: (resource, params) =>
       Promise.all(
-        params.ids.map((id) => fetchApi(UPDATE, resource, { id })),
+        params.ids.map((id) => fetchApi(UPDATE, resource, { ...params, id })),
       ).then(() => ({ data: [] })),
     create: (resource, params) => fetchApi(CREATE, resource, params),
     delete: (resource, params) => fetchApi(DELETE, resource, params),

### CODE REVIEW

**Comment by braianj** [comment]
Hello guys, when do you think you can merge this branch?

File: package.json [hunk 1]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@api-platform/admin",
-  "version": "2.2.3",
+  "version": "2.2.4",
```
alanpoulain, line 3: Could you revert this change?

alanpoulain:
Hello, thanks for the PR, it seems legit.

alanpoulain:
Are you sure this is working with only this change? Don't you need: https://github.com/api-platform/admin/pull/323?

braianj:
With that change works in my local. You also can see that with the update you are sending params but with update many you only send id. I'll push that package change to you can merge

alanpoulain:
Seems weird to me. Are you sure you're seeing `totalCommission=100` in your HTTP call? The code doesn't use the params directly to construct the search params.

braianj:
It's not about the search, it's about the PUT. When you update something in the body of the PUT is sending {} instead of {totalCommission=100}

braianj:
From my side I override the dataProvider with that one and works, also test it modifying inside the node_modules/api_platform/admin/lib/hydra/dataProvider.js and works too

alanpoulain:
OK it's the data, I thought it was the URL parameters.

alanpoulain:
Thank you!

braianj:
Thank you, awesome package

braianj:
@alanpoulain one last question, you put it on same version, right?

alanpoulain:
I'll make a release :slightly_smiling_face:

Issue #314, opened 2020-07-21
Logout Button is visible when it should not


GregoireHebert:
**API Platform version(s) affected**: 2.5.6

**Description**  
By default on first installation, on the top right burger menu the logout button is visible although we are not authenticated, in fact no authentication has been configured yet.

**How to reproduce**  
By installing the api-platform stack in 2.5.6

GregoireHebert:
Thanks @jfthuillier

PR #357, opened 2021-01-05
Do not display user menu if there is no auth provider

jfthuillier:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | https://github.com/api-platform/admin/issues/314
| License       | MIT
| Doc PR        | 

Even when no `authProvider` is configured, the user menu was displayed.

### DIFF
--- a/src/layout/AppBar.js
+++ b/src/layout/AppBar.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { AppBar } from 'react-admin';
+import { AppBar, useAuthProvider } from 'react-admin';
 import { Typography, withStyles } from '@material-ui/core';
 
 import Logo from './Logo';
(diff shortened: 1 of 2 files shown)


alanpoulain:
So we need to explicitly set it if we use authentication?

jfthuillier:
I guess so, not sure why it is not initialized here : https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/core/CoreAdminUI.tsx#L59

alanpoulain:
It is in the `AdminUI`: https://github.com/marmelab/react-admin/blob/b1a66701d79ba739bddf42b7722fac0873f53464/packages/react-admin/src/AdminUI.tsx#L17

I don't think we should disable it by default. I think React Admin has made the choice to have it by default because an admin is very often used with authentication. And we should probably do the same.

However maybe we could disable the user menu (not just the logout button) if there is no auth provider? I think it should be doable with the `useAuthProvider` hook (I think it can be used inside the `AdminResourcesGuesser` since the admin context is a parent).

jfthuillier:
I did not manage to use the hook inside the `AdminResourcesGuesser` but i have used it in the `AppBar` to set the `userMenu` value. WDYT ?

alanpoulain:
Out of curiosity, why it was not working in the `AdminResourcesGuesser`?

jfthuillier:
I did not find a property i could use at this level, that's why i had to change the behavior directly in the `AppBar`

src/layout/AppBar.js
```diff
@@ -17,7 +17,7 @@ const styles = {
 };
 
 const CustomAppBar = withStyles(styles)(({ classes, ...props }) => (
-  <AppBar {...props}>
+  <AppBar userMenu={!!useAuthProvider()} {...props}>
```
alanpoulain, line 20: Could you use the hook outside the JSX?

alanpoulain (approved):

alanpoulain:
Thank you again @jfthuillier.

PR #371, opened 2021-03-07
update react-admin to ^3.11

andrekutianski:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

update react-admin dependency to ^3.11

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "jsonld": "^3.2.0",
     "lodash.isplainobject": "^4.0.6",
     "prop-types": "^15.6.2",
-    "react-admin": "^3.0.0"
+    "react-admin": "^3.11"
   },
   "devDependencies": {
     "@babel/cli": "^7.2.3",


package.json
```diff
@@ -21,7 +21,7 @@
     "jsonld": "^3.2.0",
     "lodash.isplainobject": "^4.0.6",
     "prop-types": "^15.6.2",
-    "react-admin": "^3.0.0"
+    "react-admin": "^3.11"
```
alanpoulain, line 24: It shouldn't be needed. Why this change?

andrekutianski: react-admin 3.0.0 is from Nov 20, 2019, several corrections and bugfix was implemented since then

https://github.com/marmelab/react-admin/compare/v3.0.0...v3.11.4

alanpoulain: Sure but it just means that all the versions are supported starting from 3.0, since there is no breaking change.

dunglas:
Closing because it's not needed. The semver constraint already allows the latest version of React Admin.

Issue #387, opened 2021-06-24
Better handle file upload


ericovasconcelos:
**Description**  
The procedure described on https://marmelab.com/react-admin/CreateEdit.html#altering-the-form-values-before-submitting is not possible to do using api-platform, because it encapsulate the dataProvider. So the access to the actions to be able to include a new header according to the data being sent is not possible without breaking the api-platform dataProvider code.

**Example**  
To view the description on what I'm suggesting, please view https://marmelab.com/react-admin/CreateEdit.html#altering-the-form-values-before-

````
const dataProvider = {
    // ...
    create: (resource, params) => {
        const { **dataHeaders**, ...record } = params.data;
        const headers = new Headers({
            'Content-Type': 'application/json',
        });
        **if (dataHeaders) {       //  This can be opinated. If data contains headers array, automatically set new headers on DataProvider
            for(let i = 0; i < dataHeaders.length; i = i + 1 ) {
                headers.set(dataHeader.name,dataHeader.value);
            }         
        }**
        return httpClient(`${apiUrl}/${resource}`, {
            method: 'POST',
            body: JSON.stringify(record),
            headers,
        }).then(({ json }) => ({
            data: { ...record, id: json.id },
        }));
    },
}

**Comment by alanpoulain** [comment]
Hello,
I think it's doable by wrapping yourself the hydra data provider and by adding the headers on the `fetchHydra` parameter, like it's done for the authentication: https://api-platform.com/docs/admin/authentication-support/

**Comment by ericovasconcelos** [comment]
Hello,

Yes, actually I was able to do it by wrapping over the fetchHydra as suggested.  Thank you. 

One question: the use-case I'm using is to make the front-end to a resource that requires multipart/form-data, even when not sending the file field itself. (A resource with a upload field https://api-platform.com/docs/core/file-upload/#uploading-to-an-existing-resource-with-its-fields). 

Real-admin is smart enough to change to multipart/form-data when the file field is filled, but when the field is not filled, it falls back to json. This way, I'm overwriting the create function of DataProvider to verify if this is the case and then manually forcing the multipart:

```
const dataProvider = baseHydraDataProvider(entrypoint, fetchHydra, apiDocumentationParser, true);


const addMultipartOverrides = (dataProvider) => ({
  ...dataProvider,
  create(resource, params) {
    if (resource === 'invoices') {
      return createOperationWithMultipart(resource, params);
    } else {
      return dataProvider.create(resource, params);
    }
  }});

[...]

PR #389, opened 2021-07-10
feat: Add formEnctype support in data provider to force multipart encoding

ericovasconcelos:
| Q             | A
| ------------- | ---
| Tickets       | #387 
| License       | MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.6.5
 
-* Accept 'formEnctype: multipart' in data passed to data provider to force multipart encoding.
+* Accept 'formEnctype: multipart/form-data' in data passed to data provider to force multipart encoding.
 
 ## 2.6.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.6.5
+
+* Accept 'formEnctype: multipart' in data passed to data provider to force multipart encoding.
+
 ## 2.6.4
 
 * Call `toJSON` instead of `JSON.stringify` when there is a `toJSON` property in a form data value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.6.5
 
-* Accept 'formEnctype: multipart/form-data' in data passed to data provider to force multipart encoding.
+* Accept `extraInformation` in data passed to data provider. Use a `hasFileField` field to force multipart encoding.
 
 ## 2.6.4
 
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -345,11 +345,19 @@ export default (
         });
 
       case UPDATE:
+        const getUpdateHttpVerb = (params) => {
+          if (params.data && params.data.formMethod && params.data.formMethod === 'POST') {
+            delete params.data.formMethod;
+            return 'POST';
+          }
+          return 'PUT';
+        }
+        const updateHttpVerb = getUpdateHttpVerb(params);
         return transformReactAdminDataToRequestBody(resource, params.data).then(
           (body) => ({
             options: {
               body,
-              method: 'PUT',
+              method: updateHttpVerb,
             },
             url: itemUrl,
           }),
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -179,9 +179,7 @@ export default (
       const containFile = (element) =>
         isPlainObject(element) &&
         Object.values(element).some((value) => value instanceof File);
-      const containsMultipartOverride = (data) => {
-        return data.formEnctype === 'multipart/form-data' ? true : false;
-      };
+      const containsMultipartOverride = (data) => data.formEnctype === 'multipart/form-data';
 
       if (
         !values.some((value) => containFile(value)) &&
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -179,8 +179,19 @@ export default (
       const containFile = (element) =>
         isPlainObject(element) &&
         Object.values(element).some((value) => value instanceof File);
+      const containsMultipartOverride = (data) => {
+        if (data.formEnctype && data.formEnctype === 'multipart') {
+          delete data.formEnctype;
+          return true;
+        } else {
+          return false;
+        }
+      };
 
-      if (!values.some((value) => containFile(value))) {
+      if (
+        !values.some((value) => containFile(value)) &&
+        !containsMultipartOverride(data)
+      ) {
         return JSON.stringify(data);
       }
 
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -180,12 +180,7 @@ export default (
         isPlainObject(element) &&
         Object.values(element).some((value) => value instanceof File);
       const containsMultipartOverride = (data) => {
-        if (data.formEnctype && data.formEnctype === 'multipart/form-data') {
-          delete data.formEnctype;
-          return true;
-        } else {
-          return false;
-        }
+        return data.formEnctype === 'multipart/form-data' ? true : false;
       };
 
       if (
@@ -193,6 +188,10 @@ export default (
         !containsMultipartOverride(data)
       ) {
         return JSON.stringify(data);
+      }
+
+      if(containsMultipartOverride(data)){
+        delete data.formEnctype;
       }
 
       // If data contains a file, use FormData instead of JSON.
@@ -345,7 +344,8 @@ export default (
         });
 
       case UPDATE:
-        const updateHttpVerb = (params.data?.formMethod === 'POST') ? 'POST' : 'PUT';
+        const updateHttpVerb =
+          params.data?.formMethod === 'POST' ? 'POST' : 'PUT';
         delete params?.data?.formMethod;
         return transformReactAdminDataToRequestBody(resource, params.data).then(
           (body) => ({
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -162,10 +162,15 @@ export default (
   /**
    * @param {string} resource
    * @param {Object} data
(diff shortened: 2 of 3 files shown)


ericovasconcelos:
Hi @alanpoulain,

I´ve used 'formEnctype' and 'multipart' as the keywords to force the multipart enctype. 

The HTML specs uses:
- 'formenctype' (in lowercase) and "multipart/form-data" keyword. (https://html.spec.whatwg.org/#dom-fs-formenctype)

The API-Platform uses:
- 'multipart' and 'multipart/form-data' as keywords (https://api-platform.com/docs/core/file-upload).

Do you think it would be better to use one of the above keywords instead of 'formEnctype' and 'multipart'? It might make more sense to be in compliant with one of the standards. I was thinking that it might make more sense to keep the formEnctype and use 'multipart/form-data'. Because formEnctype is the correct IDL attribute name and 'multipart/form-data' is also compliant with HTML.  I guess that since it is running in the form processing it make sense to stick with the HTML naming. 

What do you think?

ericovasconcelos:
Hi @alanpoulain,

I´ve also added the update operation support:

- Added tests for update operation without overriding enctype and HTTP verb to POST;
- Added tests to cover the enctype and HTTP verd overriding, using 'formEnctype' and 'formMethod' keywords via data param.

Implemented the code for those scenarios on dataProvider. All tests are green. No BC.

Please see if it is ok.

src/hydra/dataProvider.js
```diff
@@ -179,8 +179,19 @@ export default (
       const containFile = (element) =>
         isPlainObject(element) &&
         Object.values(element).some((value) => value instanceof File);
+      const containsMultipartOverride = (data) => {
+        if (data.formEnctype && data.formEnctype === 'multipart') {
+          delete data.formEnctype;
```
alanpoulain, line 184: Given the name of this function, it should be pure (not mutating `data`).

src/hydra/dataProvider.js
```diff
@@ -334,11 +345,23 @@ export default (
         });
 
       case UPDATE:
+        const getUpdateHttpVerb = (params) => {
+          if (
+            params.data &&
+            params.data.formMethod &&
+            params.data.formMethod === 'POST'
+          ) {
+            delete params.data.formMethod;
```
alanpoulain, line 354: Same comment as above.

src/hydra/dataProvider.js
```diff
@@ -334,11 +345,23 @@ export default (
         });
 
       case UPDATE:
+        const getUpdateHttpVerb = (params) => {
+          if (
+            params.data &&
```
alanpoulain, line 350: Couldn't we use the optional chaining here (Babel should probably transpile it correctly)?

src/hydra/dataProvider.test.js
```diff
@@ -218,4 +218,117 @@ describe('Transform a React Admin request to an Hydra request', () => {
...
+
+    return dataProvider
+      .create('resource', {
+        data: {
+          bar: 'baz',
+          foo: 'foo',
+          formEnctype: 'multipart',
+        },
+      })
+      .then(() => {
```
alanpoulain, line 233: Why not using async/await syntax?

alanpoulain:
> Do you think it would be better to use one of the above keywords instead of 'formEnctype' and 'multipart'? It might make more sense to be compliant with one of the standards. I was thinking that it might make more sense to keep the formEnctype and use 'multipart/form-data'. Because formEnctype is the correct IDL attribute name and 'multipart/form-data' is also compliant with HTML. Since it is running in the form processing on the client-side, it makes sense to stick with the HTML naming.
>
> What do you think?

Using `multipart/form-data` seems a good idea :slightly_smiling_face:

src/hydra/dataProvider.js
ericovasconcelos: Perfect, tks.

Done.

src/hydra/dataProvider.js
ericovasconcelos: Sure. Tks.

Done.

src/hydra/dataProvider.js
ericovasconcelos: Great tip. Done.

src/hydra/dataProvider.test.js
ericovasconcelos: Sure. Updated tests to use the async/await syntax (also ajusted previous tests on dataProvider.test.js to comply).

Tks for the feedback.

ericovasconcelos: Done.

ericovasconcelos:
> > Do you think it would be better to use one of the above keywords instead of 'formEnctype' and 'multipart'? It might make more sense to be compliant with one of the standards. I was thinking that it might make more sense to keep the formEnctype and use 'multipart/form-data'. Because formEnctype is the correct IDL attribute name and 'multipart/form-data' is also compliant with HTML. Since it is running in the form processing on the client-side, it makes sense to stick with the HTML naming.
> > What do you think?
> 
> Using `multipart/form-data` seems a good idea 🙂

Dons ;)

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.6.5
+
+* Accept 'formEnctype: multipart' in data passed to data provider to force multipart encoding.
```
alanpoulain, line 5:
```suggestion
* Accept 'formEnctype: multipart/form-data' in data passed to data provider to force multipart encoding.
```

src/hydra/dataProvider.js
```diff
@@ -179,11 +179,21 @@ export default (
       const containFile = (element) =>
         isPlainObject(element) &&
         Object.values(element).some((value) => value instanceof File);
-
-      if (!values.some((value) => containFile(value))) {
+      const containsMultipartOverride = (data) => {
+        return data.formEnctype === 'multipart/form-data' ? true : false;
+      };
```
alanpoulain, line 184:
```suggestion
      const containsMultipartOverride = (data) => data.formEnctype === 'multipart/form-data';
```

src/hydra/dataProvider.js
```diff
@@ -334,11 +344,14 @@ export default (
         });
 
       case UPDATE:
+        const updateHttpVerb =
+          params.data?.formMethod === 'POST' ? 'POST' : 'PUT';
```
alanpoulain, line 348:
```suggestion
        const updateHttpMethod =
          params.data?.formMethod ?? 'PUT';
```

alanpoulain:
After some thinking, I'm not sure about the terms `formEnctype` and `formMethod`.
Since it's the data provider receiving the information, it should not be related to a "form".
I would prefer to use for instance `mediaType` and `httpMethod`.

Also, I don't really like having to delete this extra information from `data`. Ideally they should not be transferred like this but since we don't have a choice, we should make sure to not be specific to the hydra data provider.
That's why we should use a unique "extra" property in `data` used only for passing extra information to the data provider.
For instance:
```js
const transform = data => ({
    ...data, 
    extraInformation: {
        mediaType: 'multipart/form-data',
        httpMethod: 'POST',
    },
})
```
This `extraInformation` key would be then deleted once in each data provider.

And IMO we should go even further: `mediaType` and `httpMethod` are too related to the chosen data provider. We should be less specific.
For instance something like this:
```js
const transform = data => ({
    ...data, 
    extraInformation: {
        hasFile: true,
    },
})
```
And the hydra data provider then knows it has to use a multipart/form-data media type and always use the POST method.
WDYT?

ericovasconcelos:
> After some thinking, I'm not sure about the terms `formEnctype` and `formMethod`.
> Since it's the data provider receiving the information, it should not be related to a "form".
> I would prefer to use for instance `mediaType` and `httpMethod`.
> 
> Also, I don't really like having to delete this extra information from `data`. Ideally they should not be transferred like this but since we don't have a choice, we should make sure to not be specific to the hydra data provider.
> That's why we should use a unique "extra" property in `data` used only for passing extra information to the data provider.
> For instance:
> 
> ```js
> const transform = data => ({
>     ...data, 
>     extraInformation: {
>         mediaType: 'multipart/form-data',
>         httpMethod: 'POST',
>     },
> })
> ```
> 
> This `extraInformation` key would be then deleted once in each data provider.
> 
> And IMO we should go even further: `mediaType` and `httpMethod` are too related to the chosen data provider. We should be less specific.
> For instance something like this:
> 
> ```js
> const transform = data => ({
>     ...data, 
>     extraInformation: {
>         hasFile: true,
>     },
> })
> ```
> 
> And the hydra data provider then knows it has to use a multipart/form-data media type and always use the POST method.
> WDYT?

I like the idea. Makes very straightforward and clear to anyone that will use. I would just use hasFileField instead of just hasFile to make more clear that we just need the field to be present, WDYT?

May I will proceed to update the code with your suggestions?

alanpoulain:
Sure, `hasFileField` is alright.
> May I will proceed to update the code with your suggestions?
Yes you can! Sorry for changing everything after you have already worked on it.

ericovasconcelos:
Hi @alanpoulain,

I think the code looks better after your suggestions. Thanks for the feedback. I´ve done the changes and added more tests to make sure the provider totally ignores the extraInformation key (even if there is no hasFileField set to true).

alanpoulain:
Hello @ericovasconcelos. I've reworked a little your PR to delete the `extraInformation` key sooner.
I've also removed the last tests, I think it is already tested with the previous ones.
Are you OK with the changes?

ericovasconcelos (commented):
Great.

ericovasconcelos:
> Hello @ericovasconcelos. I've reworked a little your PR to delete the `extraInformation` key sooner.
> I've also removed the last tests, I think it is already tested with the previous ones.
> Are you OK with the changes?

Sure. Great job on dealing with extraInformation sooner on the code!

ericovasconcelos:
Hi @alanpoulain, is there any work left to do on this PR?

ericovasconcelos:
Hi @alanpoulain , is there any work left to do on this pull request?

alanpoulain:
Sorry for the delay, thank you @ericovasconcelos.

Issue #380, opened 2021-05-24
How to use useDataProvider


bhtomming:
I had a problem when I want to add custom data queries to the dashboard page.
In order to facilitate you to solve the problem, I'll talk about the documentation first：
.env
```
REACT_APP_API_ENTRYPOINT=https://127.0.0.1:8000/api
```
App.js
```js
import React from "react";
import { HydraAdmin,ResourceGuesser } from "@api-platform/admin";
import authProvider from "./authProvider";
import dataProvider from "./dataProvider";
import { Login,defaultTheme } from "react-admin";
import SupervisorAccountIcon from '@material-ui/icons/SupervisorAccount';
import CommentIcon from '@material-ui/icons/Comment';


import {CommentsList} from "./components/CommentsList";
import {ConferencesList} from "./components/ConferencesList";
import MyLayout from "./theme/MyLayout";
import AdminsList from "./components/AdminsList"
import GuestList from "./components/GuestList";
import Dashboard from "./theme/Dashboard";

const entrypoint = process.env.REACT_APP_API_ENTRYPOINT

const App = () => (
    <HydraAdmin
        entrypoint={ entrypoint }
        dataProvider={ dataProvider }
        authProvider={ authProvider }
        loginPage={ Login }
        theme = { defaultTheme }
        layout={MyLayout}
        dashboard={Dashboard}
    >
        <ResourceGuesser
            name={"comments"}
            list={ CommentsList }
            icon={CommentIcon}
        />
        <ResourceGuesser
            name={"conferences"}
            list={ ConferencesList }
            icon={SupervisorAccountIcon}
        />
        <ResourceGuesser
            name={"admins"}
            icon={SupervisorAccountIcon}
            list={ AdminsList }
        />
        <ResourceGuesser
            name={"guests"}
            list={ GuestList }
            icon={SupervisorAccountIcon}
        />
    </HydraAdmin>
);
export default App;


```

and the dataProvider.js
```js
import React from "react";
import { Redirect, Route } from "react-router-dom";
import { hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydra, useIntrospection } from "@api-platform/admin";
import parseHydraDocumentation from "@api-platform/api-doc-parser/lib/hydra/parseHydraDocumentation";

const entrypoint = process.env.REACT_APP_API_ENTRYPOINT;
const getHeaders = () =>
    localStorage.getItem("token")
        ? {
[...]

PR #392, opened 2021-08-18
fix(dataProvider): fix itemUrl declaration

Jisay:
| Q             | A
| ------------- | ---
| Tickets       | #380
| License       | MIT

Fix `itemUrl` in `convertReactAdminRequestToHydraRequest` in order to get, update or delete a collection item with dataProvider.

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -225,7 +225,7 @@ export default (
   const convertReactAdminRequestToHydraRequest = (type, resource, params) => {
     const entrypointUrl = new URL(entrypoint, window.location.href);
     const collectionUrl = new URL(`${entrypoint}/${resource}`, entrypointUrl);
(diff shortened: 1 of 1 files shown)


dunglas:
This fix doesn't looks good to me. The ID of the resource must be an URI, and this URI can follow any pattern. I don't get how you get and id that's not a URI in your case, but that's what need to be fixed.

Jisay:
> This fix doesn't look good to me. The ID of the resource must be an URI, and this URI can follow any pattern. I don't get how you get and id that's not a URI in your case, but that's what need to be fixed.

Thx for your answer @dunglas.

So in order to user `getOne`, `update` or `delete` method it's required to specify `id` parameter like an URI? For example:
```js
dataProvider.getOne('users', {
    id:  '/api/users/42',  // which is the full route of the resource
});
```
If so, it looks like strange to specify the full route as `id` param.

PR #400, opened 2021-09-24
fix(CreateGuesser, EditGuesser): inputChildren

f-jost:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

`inputChildren` when it is the only child is an object, putting it in an array solves the problem of using the `some` function.


alanpoulain (approved):

alanpoulain:
Thanks!

PR #402, opened 2021-10-20
fix(dataprovider): do not add properties with null value in FormData

f-jost:
| Q             | A
| ------------- | ---
| Branch?       |  main
| Tickets       | -
| License       | MIT
| Doc PR        | -

Fix `TypeError: Cannot read properties of null (reading 'toJSON')`.

When selecting an empty value from a react-admin field that allow null value like ReferenceInput, the error occurs when submitting the form that contain a file field where a FormData is used when convert react-admin data to hydra data because the `toJSON` function cannot be applied on a null value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.6.8
 
-* Make sure a property's value is not null before check if there is a `toJSON` property in a form data value
+* Make sure a property's value is not null before checking if there is a `toJSON` property in a form data value
 
 ## 2.6.7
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.6.8
+
+* Make sure a property's value is not null before check if there is a `toJSON` property in a form data value
```
alanpoulain, line 5:
```suggestion
* Make sure a property's value is not null before checking if there is a `toJSON` property in a form data value
```

alanpoulain (approved):

alanpoulain:
Thank you @f-jost!

Issue #405, opened 2021-10-29
Add possibility to disable caching in data provider


mwojtasikCodibly:
**Description**  
In actual data provider there is caching mechanizm for `getMany` requests, which may cause some mismatches when data coming from API for some resource may vary depending on actual call. The solution is to simply add flag that could disable caching mechanizm.

**Example**  
```
export default (
  entrypoint,
  httpClient = fetchHydra,
  apiDocumentationParser = parseHydraDocumentation,
  useEmbedded = false,
  disableCache = false, // added flag
) => { ... }
```

PR #406, opened 2021-10-29
feat: add possibility to disable caching in data provider

mwojtasikCodibly:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | https://github.com/api-platform/admin/issues/405 
| License       | MIT

 adds flag to data provider which will disable requests caching

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -117,6 +117,7 @@ export default (
   httpClient = fetchHydra,
   apiDocumentationParser = parseHydraDocumentation,
   useEmbedded = false, // remove this parameter for 3.0 (as true)
+  disableCache = false,
 ) => {
   /** @type {Api} */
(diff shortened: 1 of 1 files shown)


src/hydra/dataProvider.js
```diff
@@ -117,6 +117,7 @@ export default (
   httpClient = fetchHydra,
   apiDocumentationParser = parseHydraDocumentation,
   useEmbedded = false, // remove this parameter for 3.0 (as true)
+  disableCache = false,
```
alanpoulain, line 120: I think we should introduce an object for all these parameters instead of having one more parameter. WDYT?

dunglas (approved):

alanpoulain:
Superseded by https://github.com/api-platform/admin/pull/408. Thanks @mwojtasikCodibly.

PR #407, opened 2021-11-17
feat: mercure support

Shywim:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #144 
| License       | MIT
| Doc PR        | api-platform/docs#1471

Add built-in mercure support for AP's admin.

Adds a hook which can be added to any component. The hook will subscribe to the passed resource, update react-admin's store on update (using RA's actions dispatched on a successful `getOne`) and cleanup on unmount.

Adds a new optional prop to `HydraAdmin` which can be used to customize the mercure endpoint.

- [x] Show
- [x] List
- [x] Edit
- [x] Fix components tests
- [x] Doc

### DIFF
--- a/src/ListGuesser.js
+++ b/src/ListGuesser.js
@@ -33,7 +33,7 @@ const displayOverrideCode = (schema, fields) => {
   console.info(code);
 };
 
-export const DatagridBodyWithMercureSubs = (props) => {
+export const DatagridBodyWithMercure = (props) => {
   useMercureSubscription(props.resource, props.ids);
 
   return <DatagridBody {...props} />;
@@ -48,7 +48,7 @@ export const IntrospectedListGuesser = ({
   rowClick,
   rowStyle,
   isRowSelectable,
-  body = DatagridBodyWithMercureSubs,
+  body = DatagridBodyWithMercure,
   expand,
   optimized,
   children,
--- a/src/ListGuesser.js
+++ b/src/ListGuesser.js
@@ -1,6 +1,12 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import { Datagrid, List, EditButton, ShowButton, DatagridBody } from 'react-admin';
+import {
+  Datagrid,
+  List,
+  EditButton,
+  ShowButton,
+  DatagridBody,
+} from 'react-admin';
 import FieldGuesser from './FieldGuesser';
 import FilterGuesser from './FilterGuesser';
 import Introspecter from './Introspecter';
--- a/src/ListGuesser.js
+++ b/src/ListGuesser.js
@@ -1,10 +1,11 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import { Datagrid, List, EditButton, ShowButton } from 'react-admin';
+import { Datagrid, List, EditButton, ShowButton, DatagridBody } from 'react-admin';
 import FieldGuesser from './FieldGuesser';
 import FilterGuesser from './FilterGuesser';
 import Introspecter from './Introspecter';
 import Pagination from './list/Pagination';
+import useMercureSubscription from './useMercureSubscription';
 
 const displayOverrideCode = (schema, fields) => {
   if (process.env.NODE_ENV === 'production') return;
@@ -26,6 +27,12 @@ const displayOverrideCode = (schema, fields) => {
   console.info(code);
 };
 
+export const DatagridBodyWithMercureSubs = (props) => {
+  useMercureSubscription(props.resource, props.ids);
+
+  return <DatagridBody {...props} />;
+};
+
 export const IntrospectedListGuesser = ({
   fields,
   readableFields,
@@ -35,7 +42,7 @@ export const IntrospectedListGuesser = ({
   rowClick,
   rowStyle,
   isRowSelectable,
-  body,
+  body = DatagridBodyWithMercureSubs,
   expand,
   optimized,
   children,
--- a/src/ShowGuesser.test.js
+++ b/src/ShowGuesser.test.js
@@ -6,6 +6,8 @@ import FieldGuesser from './FieldGuesser';
 import { API_FIELDS_DATA } from './__fixtures__/parsedData';
 import { IntrospectedShowGuesser } from './ShowGuesser';
 
+jest.mock('./useMercureSubscription');
+
 describe('<ShowGuesser />', () => {
   test('renders with no children', () => {
     const wrapper = shallow(
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -151,7 +151,7 @@ export default (
   let apiSchema;
 
   // store mercure subscriptions
-  const subscriptions = [];
+  const subscriptions = {};
 
   /**
    * @param {Resource} resource
@@ -599,7 +599,7 @@ export default (
     subscribe: (resourceIds, callback) => {
       if (mercure !== undefined) {
         resourceIds.forEach((resourceId) => {
-          const sub = subscriptions.find((sub) => sub.id === resourceId);
+          const sub = subscriptions[resourceId];
           if (sub !== undefined) {
             sub.count++;
             return;
@@ -627,28 +627,23 @@ export default (
           };
           eventSource.addEventListener('message', eventListener);
 
-          subscriptions.push({
-            id: resourceId,
-            eventSource,
-            eventListener,
-            count: 1,
-          });
+          subscriptions[resourceId] = sub;
         });
       }
 
       return Promise.resolve({ data: null });
     },
     unsubscribe: (resource, resourceIds) => {
-      subscriptions.filter((sub) => {
-        if (resourceIds.includes(sub.id)) {
+      resourceIds.forEach((resourceId) => {
+        const sub = subscriptions[resourceId];
+        if (sub !== undefined) {
           sub.count--;
+
           if (sub.count <= 0) {
             sub.eventSource.removeEventListener('message', sub.eventListener);
-            return false;
+            delete subscriptions[resourceId];
           }
         }
-
-        return true;
       });
 
       return Promise.resolve({ data: null });
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -599,7 +599,7 @@ export default (
     subscribe: (resourceIds, callback) => {
       if (!mercure || !mercure.hub) {
         throw new Error(
-          'Mercure URL not set, did you forget to pass a `mercure` configuration to `HydraAdmin`?',
+          'Mercure hub URL not set, did you forget to pass a mercure configuration to the Hydra data provider?',
         );
       }
 
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -597,39 +597,43 @@ export default (
               );
             }),
     subscribe: (resourceIds, callback) => {
-      if (mercure !== undefined) {
-        resourceIds.forEach((resourceId) => {
-          const sub = subscriptions[resourceId];
-          if (sub !== undefined) {
-            sub.count++;
-            return;
-          }
-
-          const url = new URL(mercure.hub, window.origin);
-          url.searchParams.append(
-            'topic',
-            new URL(resourceId, mercure.topicUrl).toString(),
-          );
-
-          if (mercure.jwt !== null) {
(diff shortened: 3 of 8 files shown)


Shywim:
Maybe we could move to an object for `dataProviderFactory` arguments? If so, should we keep backward compatibility (e.g. check if the first arg is an object, if not, consider it is the way it is currently)?

alanpoulain:
> Maybe we could move to an object for `dataProviderFactory` arguments? If so, should we keep backward compatibility (e.g. check if the first arg is an object, if not, consider it is the way it is currently)?

I was talking about it in this comment, I think it's a good idea: https://github.com/api-platform/admin/pull/406#discussion_r742833588

Yes you should keep the BC.

src/ListGuesser.js
```diff
@@ -26,6 +33,12 @@ const displayOverrideCode = (schema, fields) => {
   console.info(code);
 };
 
+export const DatagridBodyWithMercureSubs = (props) => {
```
alanpoulain, line 36:
```suggestion
export const DatagridBodyWithMercure = (props) => {
```

To be simpler?

src/hydra/dataProvider.js
```diff
@@ -564,5 +568,53 @@ export default (
                   (status ? `Status: ${status}` : ''),
               );
             }),
+    subscribe: (resourceIDs, callback) => {
```
alanpoulain, line 571:
```suggestion
    subscribe: (resourceIds, callback) => {
```

src/hydra/dataProvider.js
```diff
@@ -564,5 +568,53 @@ export default (
...
+      resourceIDs.forEach((resourceID) => {
```
alanpoulain, line 572:
```suggestion
      resourceIDs.forEach((resourceId) => {
```

src/hydra/dataProvider.js
```diff
@@ -564,5 +568,53 @@ export default (
...
+        const sub = subscriptions.find((sub) => sub.id === resourceID);
+        if (sub != null) {
```
alanpoulain, line 574:
```suggestion
        if (sub !== null) {
```

src/hydra/dataProvider.js
```diff
@@ -564,5 +568,53 @@ export default (
...
+        url.searchParams.append(
+          'topic',
+          new URL(resourceID, entrypoint).toString(),
+        );
+        const eventSource = new EventSource(url.toString());
+        const eventListener = (event) => {
+          const document = transformJsonLdDocumentToReactAdminDocument(
+            JSON.parse(event.data),
+          );
+          // we need redux's `dispatch` from the react tree
```
alanpoulain, line 589: I don't understand this comment.

src/hydra/dataProvider.js
```diff
@@ -564,5 +568,53 @@ export default (
...
+          id: resourceID,
+          eventSource,
+          eventListener,
+          count: 1,
+        });
+      });
+
+      return Promise.resolve({ data: null });
+    },
+    unsubscribe: (resource, resourceIDs) => {
```
alanpoulain, line 604:
```suggestion
    unsubscribe: (resource, resourceIds) => {
```

alanpoulain:
For information, I thought it would have been great to retrieve the Mercure's hub link from the response headers, but sadly it's difficult to do so. See https://github.com/marmelab/react-admin/pull/4834.

alanpoulain:
Also, the ra-realtime API Platform adapter has two other parameters for the data provider, I think we could introduce them too: https://marmelab.com/ra-enterprise/modules/ra-realtime#api-platform-adapter

src/hydra/dataProvider.js
Shywim: The callback to the hook is necessary only because we need to dispatch an action, if we could dispatch from the `dataProvider`, we could do everything from withing `subscribe`.

I tried to word it better

src/ShowGuesser.test.js
```diff
@@ -6,6 +6,8 @@ import FieldGuesser from './FieldGuesser';
 import { API_FIELDS_DATA } from './__fixtures__/parsedData';
 import { IntrospectedShowGuesser } from './ShowGuesser';
 
+jest.mock('./useMercureSubscription');
```
Shywim, line 9: Inclusion of `useMercureSubscription` bring in a dependency on redux. Problem is, react-redux v6+ [doesn't play nicely with enzyme](https://github.com/enzymejs/enzyme/issues/2282)

What I've tried:
```js
const DummyReduxProvider = ({ children }) => (
  <Provider store={createStore(() => {})}>{children}</Provider>
);

const wrapper = shallow(
  <IntrospectedShowGuesser
    resource="user"
    schema={{ name: 'users', title: 'User' }}
    readableFields={API_FIELDS_DATA}
    id="ShowComponentId"
  />,
  {
    wrappingComponent: DummyReduxProvider,
  },
);
```

src/hydra/dataProvider.js
```diff
@@ -113,14 +119,36 @@ export const transformJsonLdDocumentToReactAdminDocument = (
...
 export default (
-  entrypoint,
+  entrypointOrParams,
   httpClient = fetchHydra,
   apiDocumentationParser = parseHydraDocumentation,
   useEmbedded = false, // remove this parameter for 3.0 (as true)
 ) => {
+  let mercure;
+  let entrypoint = entrypointOrParams;
+  if (typeof entrypointOrParams === 'object') {
```
alanpoulain, line 129: We should show a deprecated message if it's not the case.

src/hydra/dataProvider.js
```diff
@@ -113,14 +119,36 @@ export const transformJsonLdDocumentToReactAdminDocument = (
...
+    const params = Object.assign({}, defaultParams, entrypointOrParams);
```
alanpoulain, line 130: Why not using the spread operator?

src/hydra/dataProvider.js
```diff
@@ -564,5 +592,62 @@ export default (
                   (status ? `Status: ${status}` : ''),
               );
             }),
+    subscribe: (resourceIds, callback) => {
+      if (mercure !== undefined) {
```
alanpoulain, line 596: Error if it's not the case?

src/hydra/dataProvider.test.js
```diff
@@ -112,11 +112,12 @@ describe('Transform a React Admin request to an Hydra request', () => {
...
     }),
   );
-  const dataProvider = dataProviderFactory(
-    'entrypoint',
-    mockFetchHydra,
-    mockApiDocumentationParser,
-  );
+  const dataProvider = dataProviderFactory({
+    entrypoint: 'entrypoint',
+    mercureHub: 'entrypoint',
```
alanpoulain, line 117: `mecure` object?

src/useMercureSubscription.js
```diff
@@ -0,0 +1,34 @@
...
+import { useEffect } from 'react';
+import { useDispatch } from 'react-redux';
+
+export default function useMercureSubscription(resource, idOrIds) {
+  const dataProvider = useDataProvider();
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    const ids = Array.isArray(idOrIds) ? idOrIds : [idOrIds];
+    dataProvider.subscribe(ids, (document) => {
```
alanpoulain, line 16: I think it could be nice to verify that the data provider used has the `subscribe` and `unsubscribe` methods, and log a message if it's not the case.
It will allow to use other data providers than the Hydra one.

src/hydra/dataProvider.js
```diff
@@ -113,14 +119,36 @@ export const transformJsonLdDocumentToReactAdminDocument = (
...
+      },
+      params.mercure,
+    );
+  }
+
   /** @type {Api} */
   let apiSchema;
 
+  // store mercure subscriptions
+  const subscriptions = [];
```
alanpoulain, line 150: I think it would be better to use an object here, with the resource ids as keys.

src/hydra/dataProvider.js
```diff
@@ -564,5 +596,61 @@ export default (
...
+    },
+    unsubscribe: (resource, resourceIds) => {
+      resourceIds.forEach((resourceId) => {
+        const sub = subscriptions[resourceId];
+        if (sub !== undefined) {
+          sub.count--;
+
+          if (sub.count <= 0) {
+            sub.eventSource.removeEventListener('message', sub.eventListener);
+            delete subscriptions[resourceId];
```
Shywim, line 648: I used `delete` here by default, but I know it is slower than setting to `undefined`, so I can swap if it's preferred :slightly_smiling_face:

alanpoulain: I don't think we need micro-optimizations like this. It's alright like this.

src/hydra/dataProvider.js
```diff
@@ -113,14 +119,40 @@ export const transformJsonLdDocumentToReactAdminDocument = (
...
+    useEmbedded = params.useEmbedded;
+    mercure = {
+      hub: `${params.entrypoint}/.well-known/mercure`,
+      jwt: null,
+      topicUrl: window.origin,
+      ...params.mercure,
+    };
+  } else {
+    console.warn(
+      'Passing a list of arguments is deprecated! Use an object instead.',
```
alanpoulain, line 146:
```suggestion
      'Passing a list of arguments for building the data provider is deprecated. Please use an object instead.',
```

src/hydra/dataProvider.js
```diff
@@ -564,5 +596,61 @@ export default (
                   (status ? `Status: ${status}` : ''),
               );
             }),
+    subscribe: (resourceIds, callback) => {
+      if (!mercure || !mercure.hub) {
+        throw new Error(
+          'Mercure URL not set, did you forget to pass a `mercure` configuration to `HydraAdmin`?',
```
alanpoulain, line 602:
```suggestion
          'Mercure hub URL not set, did you forget to pass a mercure configuration to the Hydra data provider?',
```

src/hydra/dataProvider.js
```diff
@@ -564,5 +596,61 @@ export default (
...
+
+        subscriptions[resourceId] = sub;
+      });
+
+      return Promise.resolve({ data: null });
+    },
+    unsubscribe: (resource, resourceIds) => {
+      resourceIds.forEach((resourceId) => {
+        const sub = subscriptions[resourceId];
+        if (sub !== undefined) {
```
alanpoulain, line 643: Could you use an early return instead?

src/useMercureSubscription.js
```diff
@@ -0,0 +1,45 @@
...
+
+  useEffect(() => {
+    const ids = Array.isArray(idOrIds) ? idOrIds : [idOrIds];
+
+    if (
+      dataProvider.subscribe === undefined ||
+      dataProvider.unsubscribe === undefined
+    ) {
+      console.warn(
+        '`subscribe` and/or `unsubcribe` methods were not set in the `dataProvider`, mercure realtime update functionnalities will not work',
```
alanpoulain, line 22:
```suggestion
        'subscribe and/or unsubscribe methods were not set in the data provider, Mercure realtime update functionalities will not work. Please use a compatible data provider.',
```

src/useMercureSubscription.js
```diff
@@ -0,0 +1,45 @@
...
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    const ids = Array.isArray(idOrIds) ? idOrIds : [idOrIds];
+
+    if (
+      dataProvider.subscribe === undefined ||
+      dataProvider.unsubscribe === undefined
+    ) {
+      console.warn(
```
alanpoulain, line 21: This console message will probably show a lot. Maybe we should use a ref to show it only once?

src/useMercureSubscription.js
```diff
@@ -0,0 +1,49 @@
...
+  const didShowNoSubscribeMethodWarning = useRef(false);
+
+  useEffect(() => {
+    const ids = Array.isArray(idOrIds) ? idOrIds : [idOrIds];
+
+    if (
+      (!didShowNoSubscribeMethodWarning.current &&
+        dataProvider.subscribe === undefined) ||
+      dataProvider.unsubscribe === undefined
+    ) {
```
alanpoulain, line 23:
```suggestion
    if (
      !didShowNoSubscribeMethodWarning.current &&
        (dataProvider.subscribe === undefined ||
      dataProvider.unsubscribe === undefined)
    ) {
```

alanpoulain:
:tada: Thank you very much @Shywim!

Issue #259, opened 2019-11-15
[BUG] object id being reassigned as iri is breaking edits


jspizziri:
**Admin version(s) affected**: 1.0.2

**Description**  
The id reassignment of `@id` to `id` in [the constructor of the ReactAdminDocument](https://github.com/api-platform/admin/blob/master/src/hydra/dataProvider.js#L17) is resulting in edits being impossible. The issue is that this constructor results in mapping this:

```json
{
  "@context": "/contexts/Foo",
  ​​"@id": "/api/foos/1",
  "@type": "Foo",
  "id": 1,
  "​​name": "Bar Baz"
}
```

To this:
```json
{
  "@context": "/contexts/Foo",
  ​​"@id": "/api/foos/1",
  "@type": "Foo",
  "id": "/api/foos/1",
  "originId": 1,
  "​​name": "Bar Baz"
}
```

Subsequently, when the EditGuesser, and InputGuesser construct the input fields for this document, it correctly assess the type of the `id` field as a `number`, but then sets the value to the IRI string. So you get the following:

```html
<input 
  aria-invalid="false" 
  class="MuiInput-input-272 MuiInput-inputType-275" 
  id="id" 
  name="id" 
  type="number" 
  value="/api/foos/1"
>
```

Firstly this breaks the input, as no id actually shows up in the UI (number inputs can't parse strings), and secondly when attempting to save the form you get an error from the API saying there's a type mismatch, as it's expecting an int id, but received a string.

**How to reproduce**  
Spin up the api-platform, and attempt to edit a document in admin which has an integer id.

**Possible Solution**  
The obvious option seems to be to not reassign the id field; however, I'm assuming that was implemented for a reason, and its removal would cause regressions elsewhere. Apart from that it would probably be required to add some conditional logic to the `InputGuesser` so that `if (field.name === 'id') { value = field.originId }`.

**Additional Context**  
You can work around this issue by manually copying the resources id and entering into the id field, then saving. However, that's clearly not a sustainable solution.

Moutard3:
Got a problem similar with a Uuid typed Id

```json
{
  "@context": "/contexts/User",
  "@id": "/users/002c9895-ecea-4f50-8d01-e6634cfaba17",
  "@type": "User",
  "email": "user@email.tld",
  "originId": null,
  "id": "/users/002c9895-ecea-4f50-8d01-e6634cfaba17",
  "password": "password"
}
```

`id` can't be converted to Uuid and throw an error.

[...]

PR #411, opened 2021-12-22
use original id for update instead of IRI

Perni1984:
| Q             | A
| ------------- | ---
| Branch?       |  main
| Tickets       | #259 
| License       | MIT
| Doc PR        | n/a

fixes https://github.com/api-platform/admin/issues/259

### DIFF
--- a/src/hydra/dataProvider.js
+++ b/src/hydra/dataProvider.js
@@ -453,9 +453,7 @@ export default (
         const updateHttpMethod = extraInformation.hasFileField ? 'POST' : 'PUT';
 
         // fixes https://github.com/api-platform/admin/issues/259
-        if (params && params.data && params.data.id && params.originId) {
-          params.data.id = params.originId;
-        }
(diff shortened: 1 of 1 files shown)


src/hydra/dataProvider.js
```diff
@@ -451,6 +451,10 @@ export default (
 
       case UPDATE:
         const updateHttpMethod = extraInformation.hasFileField ? 'POST' : 'PUT';
+
+        // fixes https://github.com/api-platform/admin/issues/259
+        params.data.id = params.originId;
```
alanpoulain, line 456: I think it can be dangerous to modify the params by reference. Maybe it would be better to clone it, don't you think?

Perni1984: @alanpoulain you're right, I changed it to copy the values.

dunglas (approved):

alanpoulain:
Superseded by https://github.com/api-platform/admin/pull/427.

Issue #230, opened 2019-10-01
Migrate the library to TypeScript


dunglas:
As React Admin migrated to TypeScript, we should migrate API Platform Admin too. PropTypes definitions must be kept for consumer applications not written in TS.

instalab:
I think I'll claim this task if possible.
I've been working on a bigger JS React projects at work and would like to learn TypeScript now.

instalab:
@dunglas Any additional notes you'd like to add, feel free ;)

instalab:
OK just a note from myself, the guys from React Admin haven't migrated fully yet.

instalab:
@dunglas Do you have Slack?
Could I join if you do?

dunglas:
@instalab We do! https://api-platform.com/support/

pierre-H:
Any news ? 😃

dunglas:
@pierre-H AFAIK no one is currently working on this.

Fabious:
I'm working on it, expect a PR soon 😉

PR #412, opened 2021-12-27
feat: Typescript migration

Fabious:
Here comes TypeScript!

Somes remarks:
- i avoid touching the code implementation as much as possible
- I left some `any` and interface with lots of any that will need update in the future for better typing (as it needs more dedication)
- some `TODOs`

I have tested it locally with a base api-platform instance. If someone wants to help to test it against a real API maybe?

edit: tested with api-platform/demo, seems all good!

Related: #230


src/hydra/fetchHydra.ts
```diff
@@ -0,0 +1,85 @@
+import { HttpError } from 'react-admin';
```
lutangar, line 1: You probably should `git mv` those files instead of deleting and recreating them. This would preserve git history.

package.json
```diff
@@ -29,6 +29,10 @@
     "@babel/plugin-proposal-export-default-from": "^7.2.0",
```
alanpoulain, line 29: Should be removed as well.

package.json
```diff
@@ -47,20 +51,22 @@
...
   "scripts": {
-    "build": "babel src -d lib --ignore '*.test.js'",
+    "build": "babel src -d lib --extensions .ts,.tsx --ignore '*.test.js'",
     "eslint-check": "eslint-config-prettier .eslintrc.js",
-    "fix": "eslint --ignore-pattern 'lib/*' --ext .js,.md --fix .",
-    "lint": "eslint --ignore-pattern 'lib/*' --ext .js,.md .",
+    "fix": "eslint --ignore-pattern 'lib/*' --ext .ts,.tsx,.js,.md --fix .",
+    "lint": "eslint --ignore-pattern 'lib/*' --ext .ts,.tsx,.js,.md .",
     "test": "jest src",
+    "test:types": "tsc",
```
alanpoulain, line 67: This is not really a command in the `test` namespace IMO.
Maybe a `compile` command instead?

Fabious: I was inspired by react-query for that name

Hmm it's not really compiling, more like a check. Maybe something like check:ts or check:types ?

Naming is hard!

src/hydra/fetchHydra.ts
Fabious: Done (i used 1 commit to revert back to js, then 1 commit to rename to ts/tsx using `git mv`)

package.json
alanpoulain: It seems it is the choice of Next.js too: https://github.com/vercel/next.js/blob/34dcede6902fc1bc100a3a9d8874851c4bcd9977/package.json#L14.
OK then!

prettier.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  printWidth: 80,
```
alanpoulain, line 2: Why not removing the default values?

dunglas:
Is it necessary to keep Babel? Can't we use the typescript compiler directly?

src/AdminGuesser.tsx
```diff
@@ -17,6 +19,25 @@ import ResourceGuesser from './ResourceGuesser';
...
+  dataProvider: CustomDataProvider;
+  schemaAnalyzer: any;
+  includeDeprecated: boolean;
+}
+
+interface AdminGuesserWithErrorProps extends AdminGuesserProps {
+  error: string;
+}
+
+interface AdminRessourcesGuessProps extends Omit<AdminProps, 'loading'> {
```
alanpoulain, line 34:
```suggestion
interface AdminRessourcesGuesserProps extends Omit<AdminProps, 'loading'> {
```

src/CreateGuesser.tsx
```diff
@@ -31,9 +32,27 @@ const displayOverrideCode = (schema, fields) => {
...
+  toolbar: any;
+  validate: any;
+  variant: any;
+  warnWhenUnsavedChanges: any;
+  writableFields: any;
+}
+
 export const IntrospectedCreateGuesser = ({
-  fields,
-  readableFields,
```
alanpoulain, line 36: Are you sure it's a good idea to remove them? We probably want to filter them.

src/EditGuesser.tsx
```diff
@@ -32,9 +33,25 @@ const displayOverrideCode = (schema, fields) => {
...
+  toolbar: any;
+  validate: any;
+  variant: any;
+  warnWhenUnsavedChanges: any;
+  writableFields: any;
+}
+
 export const IntrospectedEditGuesser = ({
-  fields,
-  readableFields,
```
alanpoulain, line 37: Are you sure it's a good idea to remove them? We probably want to filter them.

src/FieldGuesser.tsx
```diff
@@ -88,8 +88,6 @@ const renderField = (field, schemaAnalyzer, props) => {
 
 export const IntrospectedFieldGuesser = ({
   fields,
-  readableFields,
-  writableFields,
```
alanpoulain, line 92: Are you sure it's a good idea to remove them? We probably want to filter them.

src/hydra/dataProvider.ts
```diff
@@ -662,14 +657,15 @@ export default (
       apiSchema
         ? Promise.resolve({ data: apiSchema })
         : apiDocumentationParser(entrypoint)
-            .then(({ api, customRoutes = [] }) => {
```
alanpoulain, line 665: Why removing the `customRoutes`?

src/types.ts
```diff
@@ -0,0 +1,15 @@
+import { DataProvider, CustomRoutes } from 'react-admin';
+
+export interface CustomDataProvider extends DataProvider {
```
alanpoulain, line 3:
```suggestion
export interface ApiPlatformAdminDataProvider extends DataProvider {
```

alanpoulain:
It would be nice to generate the types for the exported lib (and the source maps too).

src/CreateGuesser.tsx
Fabious: They are not used (`'fields' is defined but never used. eslint@typescript-eslint/no-unused-vars`)

Do you prefer that i put it back and modify the eslint rules or leave as it ? (same for the other comments)

src/hydra/dataProvider.ts
Fabious: good catch, i did this by error because of typescript infering the wrong type on this one and raising an error (`Property 'customRoutes' does not exist on type '{ api: Api; response: Response; status: number; }'.ts(2339)`) 

I revert it to the original with an any type for the moment

Fabious:
@dunglas yes we still need babel as i haven't migrated the tests (next PR?), but eventually i can look to make the switch

@alanpoulain yes good idea =] maybe in an other PRs? i need to look the cleanest way to do it (maybe it's easy? don't really know as i don't work often with libs)

src/CreateGuesser.tsx
alanpoulain: React-admin is using a `sanitizeRestProps` function, maybe we can use something like this as well?

prettier.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  tabWidth: 2,
```
alanpoulain, line 2: There are still some rules with the default values (tabWidth, useTabs...).

src/AdminGuesser.tsx
```diff
@@ -17,6 +19,25 @@ import ResourceGuesser from './ResourceGuesser';
...
+  dataProvider: ApiPlatformAdminDataProvider;
+  schemaAnalyzer: any;
+  includeDeprecated: boolean;
+}
+
+interface AdminGuesserWithErrorProps extends AdminGuesserProps {
+  error: string;
+}
+
+interface AdminRessourcesGuesserProps extends Omit<AdminProps, 'loading'> {
```
alanpoulain, line 34:
```suggestion
interface AdminResourcesGuesserProps extends Omit<AdminProps, 'loading'> {
```

src/Introspecter.tsx
```diff
@@ -142,7 +166,6 @@ const Introspecter = ({
 
 Introspecter.propTypes = {
   component: PropTypes.elementType.isRequired,
-  resource: PropTypes.string.isRequired,
```
alanpoulain, line 145: Is it not required?

Fabious: There was an error also seen in last PR 411 

https://github.com/api-platform/admin/runs/4630829107

Fabious: Maybe putting it back but not required? Like that? `resource: PropTypes.string`

alanpoulain: Maybe it's really an issue when `FieldGuesser` is used. It should not work without a related resource.

alanpoulain: I think it is required though. We should check it!

Fabious: Found the culprit, but it's weird because it is a test written 2 years ago. Can you check it? If you can at my last commit i added a `test.skip`.

Clearly not ideal, so i don't know if we patch this now or create an issue for later fix.

alanpoulain (approved):

alanpoulain:
Very good start for TypeScript support!
Thank you @Fabious.

PR #416, opened 2022-01-09
refactor: move tests to their own folders

Fabious:
Proposition to move tests to their own folders

Maybe it's a personal preference but i find it clearer for this project to make a separation. Especially moving __fixtures__ and __snapshots__ in a dedicated folder feels nice.

Take or leave :D


Fabious:
Also fixed babel compiling tests files

alanpoulain:
Personally, I think it's better to have test files next to the related components. And it's the common practice in the React world.
However why not for the fixtures and the snapshots.

Fabious:
Agree in React we often colocate files (like Kent said: https://kentcdodds.com/blog/colocation#tests)

However `src` could use some organization, maybe having folders like `hooks`, `contexts`, `utils` or even `core` (for the Guessers). Mimicking react-admin maybe?

Grouping stuffs together to have better colocation =]
WDYT?

I guess im overthinking this stuff :D just throwing ideas, i will close this but discussion is open =]

alanpoulain:
Yes maybe it would be better indeed :slightly_smiling_face:

PR #420, opened 2022-01-12
fix: use pagination for get many items with id filter

mRoca:
When using a ReferenceField into a list, and when the related entity has an `id[]` filter, the list is currently displaying empty cells if we want to display more lines than the related entity perPage default value set on the API config.

```jsx
  <ListGuesser {...props} perPage={100}>
    ...
    {/* If the User resource's default itemsPerPage value is less than 100, this component can display an empty cell */}
    <ReferenceField source="targetUser" reference="users" link="show">
      <TextField source="fullname" />
    </ReferenceField>
    ...
  </ListGuesser>
```

### DIFF
--- a/src/hydra/dataProvider.ts
+++ b/src/hydra/dataProvider.ts
@@ -625,7 +625,7 @@ function dataProvider(
         if (result) {
           return fetchApi(GET_LIST, resource, {
             pagination: {
-              // Asking for the good amount of items, as we could want to retrieve more items than the default itemsPerPage value.
+              // Asking for the good amount of items, as we could want to retrieve more items than the default amount sent by the API.
               perPage: params.ids?.length,
             },
             sort: {},
--- a/src/hydra/dataProvider.ts
+++ b/src/hydra/dataProvider.ts
@@ -5,6 +5,9 @@ import {
   GET_MANY_REFERENCE,
   GET_ONE,
   UPDATE,
+  GetListResult,
+  GetListParams,
+  GetManyReferenceParams,
 } from 'react-admin';
 import isPlainObject from 'lodash.isplainobject';
 import { parseHydraDocumentation } from '@api-platform/api-doc-parser';
@@ -376,7 +379,9 @@ function dataProvider(
           sort: { field, order },
         } = params;
 
-        if (order) collectionUrl.searchParams.set(`order[${field}]`, order);
+        if (order && field) {
+          collectionUrl.searchParams.set(`order[${field}]`, order);
+        }
         if (page) collectionUrl.searchParams.set('page', page);
         if (perPage) collectionUrl.searchParams.set('itemsPerPage', perPage);
         if (params.filter) {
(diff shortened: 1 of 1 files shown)


src/hydra/dataProvider.ts
```diff
@@ -624,7 +624,10 @@ function dataProvider(
         // Hydra doesn't handle MANY requests but if a search filter for the id is available, it is used.
         if (result) {
           return fetchApi(GET_LIST, resource, {
-            pagination: {},
+            pagination: {
+              // Asking for the good amount of items, as we could want to retrieve more items than the default itemsPerPage value.
```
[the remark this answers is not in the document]
mRoca: The default value is only set on the API side, and as no itemsPerPage value is currently sent, the limit is the one you set on your sf project

alanpoulain, line 628:
```suggestion
              // Asking for the good amount of items, as we could want to retrieve more items than the default amount sent by the API.
```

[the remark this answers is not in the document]
alanpoulain: I've made a suggestion to make it clearer then.

src/hydra/dataProvider.ts
```diff
@@ -624,7 +624,10 @@ function dataProvider(
...
+              perPage: params.ids?.length,
```
alanpoulain, line 629: Can't we use the `perPage` param from the `getMany` method?

mRoca:
Ok, I've checked the `react-admin` source code: the `getMany()` method should not deal with pagination.

So we have 2 options:
- setting the `itemsPerPage` value, but the issue will still be present if the `ids.length` value is greater than the `max_items_per_page` value on the API side
- running as many requests as necessary and return a value containing all merged pages, but it's a lot more of work

src/hydra/dataProvider.ts
mRoca: I don't understand: no pagination param are sent to the `getMany` method

alanpoulain:
> running as many requests as necessary and return a value containing all merged pages, but it's a lot more of work

I think this is the right solution. However I don't think we can know in advance the value of `max_items_per_page` (not sent in Hydra metadata), so the requests need to be run sequentially. Hopefully the case will be very occasional.

mRoca:
@alanpoulain I've added a new commit with a `fetchAllPages` new function. 2 points:
- I've used TS types for the fist time in this file, is you plan to migrate everything to TS types, or should we keep the old JSdoc ?
- the function is using react-admin data format, so the Hydra pagination is not well used. It's possible to improve it by using the Hydra doc, but I'll have to rewrite a lot of code. imho the PR could be merged as it, and the improvement can be done later if needed.

src/hydra/dataProvider.ts
```diff
@@ -604,6 +609,51 @@ function dataProvider(
...
+   */
+  const fetchAllPages = async (
+    resource: string,
+    params: Partial<GetListParams>,
+    previousResult?: GetListResult,
+  ): Promise<GetManyResult> => {
+    params.pagination = {
+      // Using a default perPage value big enough to limit the number of requests
+      // If the API has a max_per_page value, the perPage parameter can be ignored
+      perPage: params.pagination?.perPage || 100,
```
alanpoulain, line 624: I don't think we should set a default value when there is no `perPage` parameter.

src/hydra/dataProvider.ts
```diff
@@ -604,6 +609,51 @@ function dataProvider(
...
+      page: params.pagination?.page || 1,
+    };
+
+    // Get the current page data
+    const pageResult = (await fetchApi(GET_LIST, resource, {
```
alanpoulain, line 629: `GET_LIST` could be a parameter (it could be used with `GET_MANY_REFERENCE` for instance).

src/hydra/dataProvider.ts
```diff
@@ -604,6 +609,51 @@ function dataProvider(
...
+      ...params,
+      sort: { ...(params.sort || {}) }, // GetListParams requires a full 'sort' object, but it's not mandatory for the Hydra method
+    })) as GetListResult; // As the fetchApi return is not typed for now, assuming that the result is the one we want
+
+    const result = previousResult ?? pageResult;
```
alanpoulain, line 634: Why not:
```ts
const result = [...(previousResult ?? []), ...pageResult];
```

alanpoulain:
> I've used TS types for the fist time in this file, is you plan to migrate everything to TS types, or should we keep the old JSdoc ?

We just have migrated to TS, and we plan to migrate everything to TS (and remove the JSDoc).

src/hydra/dataProvider.ts
mRoca: No, we really want to run the GET_LIST operation here, as it's the only one returning a data AND a total number of items.

src/hydra/dataProvider.ts
mRoca: If you look at the `GetListParams` type, you can see that we must provide both perPage & page parameters with a number: they are not optional for react-admin.

Indeed, the hydra provider does not requires them for now, but I guess you are going to type everything with react-admin types soon.

src/hydra/dataProvider.ts
alanpoulain: It seems to me that is also the case with `GetManyReferenceResult`.

src/hydra/dataProvider.ts
mRoca: previousResult & pageResult are not arrays, but objects with a data & total keys.

src/hydra/dataProvider.ts
alanpoulain: Alright.

src/hydra/dataProvider.ts
mRoca: Ok, I can add allow both GET_LIST and GET_MANY_REFERENCE

mRoca: `export const fetchActionsWithTotalResponse = ['getList', 'getManyReference'];` => perfect :-)

mRoca: Done, can you check if it's okayfor you ?

alanpoulain (approved):

alanpoulain:
Thank you @mRoca.

PR #421, opened 2022-01-19
feat(hydra): migrate to full TS

alanpoulain:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A


src/hydra/dataProvider.ts
```diff
@@ -268,22 +299,33 @@ function dataProvider(
...
+    ).then((data) => {
       const values = Object.values(data);
-      const containFile = (element) =>
+      const containFile = (element: unknown) =>
         isPlainObject(element) &&
-        Object.values(element).some((value) => value instanceof File);
+        Object.values(element as { [key: string]: unknown }).some(
+          (value) => value instanceof File,
+        );
+      type toJSONObject = { toJSON(): string };
```
soyuka, line 322: move to types?

src/hydra/dataProvider.ts
```diff
@@ -293,15 +335,17 @@ function dataProvider(
       }
 
       const body = new FormData();
-      Object.entries(data).map(([key, value]: [key: string, value: any]) => {
+      Object.entries<string | { toJSON(): string }>(
```
soyuka, line 338: reuse the tojson type

src/hydra/dataProvider.ts
```diff
@@ -354,52 +391,63 @@ function dataProvider(
...
+        if (perPage) url.searchParams.set('itemsPerPage', perPage.toString());
+        if (filter) {
+          const buildFilterParams = (
+            key: string,
+            nestedFilter: {
+              [key: string]:
+                | string
+                | boolean
+                | number
+                | { [key: string]: string | boolean | number };
```
soyuka, line 433: this is a valid case for `any` imo

src/types.ts
```diff
@@ -1,15 +1,277 @@
...
+  mercure?: Partial<MercureOptions>;
+  useEmbedded?: boolean;
+  disableCache?: boolean;
+}
 
 export interface ApiPlatformAdminDataProvider extends DataProvider {
+  getList: <RecordType extends ApiPlatformAdminRecord>(
+    resource: string,
+    params: ApiPlatformAdminGetListParams,
+  ) => Promise<GetListResult<RecordType>>;
```
soyuka, line 224: you could create a type to avoid repeating all this, for example:

`type ApiPlatformDataProviderType<T> = RecordType extends ApiPlatformAdminRecord>(
    resource: string,
    params: ApiPlatformAdminGetListParams,
  ) => Promise<T<RecordType>>'

```
getList: ApiPlatformDataProviderType<GetListResult>
```

src/types.ts
```diff
@@ -1,15 +1,277 @@
...
   }>;
-  subscribe: (...args: any) => void;
+  subscribe: (
+    resourceIds: string[],
+    callback: (document: ApiPlatformAdminRecord) => void,
+  ) => Promise<{ data: null }>;
+  unsubscribe: (
+    _resource: string,
+    resourceIds: string[],
+  ) => Promise<{ data: null }>;
```
soyuka, line 271: why not use `Promise<void>`?

src/types.ts
```diff
@@ -1,15 +1,277 @@
...
 }
+
+export type FilterParameter = {
+  name: string;
+  isRequired: boolean;
+};
```
soyuka, line 277: on a general notes having types and interfaces in the same file is confusing

src/hydra/dataProvider.ts
```diff
@@ -268,22 +299,33 @@ function dataProvider(
 
   const transformReactAdminDataToRequestBody = (
     resource: string,
-    data: object,
-    extraInformation: { hasFileField?: any },
-  ): Promise<any> => {
-    /** @type {Resource} */
+    data: { [key: string]: unknown } | unknown,
```
Fabious, line 302: equivalent to `data: unknown`

(that's why you need to cast type, L314)

Fabious: maybe we need to check the type of data to avoid casting (`data as FormData`, `data as { [key: string]: unknown }`)

src/hydra/dataProvider.ts
```diff
@@ -5,28 +5,59 @@ import {
...
+  DataProviderType,
+  HydraCollection,
+  HydraDataProviderFactoryParams,
+  HydraHttpClientResponse,
+  MercureOptions,
+  MercureSubscription,
+  SearchParams,
+} from '../types';
+
+const isPlainObject = (value: any): value is object =>
```
Fabious, line 43: wrapper needed?

src/hydra/dataProvider.ts
```diff
@@ -222,44 +248,49 @@ function dataProvider(
...
+      });
+    }
 
-    const fieldDataKeys: any = Object.keys(fieldData);
+    const fieldDataKeys = Object.keys(fieldData);
     const fieldDataValues = Object.values(fieldData);
 
     return Promise.all(fieldDataValues).then((fieldData) => {
-      const object = {};
+      const object: { [key: string]: unknown } = {};
```
Fabious, line 288:
```suggestion
      const object: Record<string: unknown> = {};
```
Apply to everywhere you have `{ [key: string]: something }` (`Record<string, something>`)

src/hydra/dataProvider.ts
Fabious: `Record<string, unknown>`  maybe

src/hydra/dataProvider.ts
alanpoulain: This is really an internal type, I think it can stay here.

src/hydra/dataProvider.ts
alanpoulain: Filter values have only these types: it's useful to use the exact types.

src/types.ts
alanpoulain: It's trickier than that: the result needs to use a generic type so it can't be easily replaced.

src/types.ts
alanpoulain: Because React-Admin data provider methods need to return something in order to be considered successful.

src/hydra/dataProvider.ts
alanpoulain: We cannot really check it, we don't really know the type of `data` when `apiResource` is not found.

Issue #413, opened 2022-01-06
Hangs/Crashes/Unresponsiveness after a few sort and/or filtering actions


4c0n:
**API Platform version(s) affected**: 2.6.7

**Description**  
When using filters for filtering or sorting together with Mercure. After a few clicks things get really slow and/or unresponsive.
Eventually the eventsource type requests fail and allow for a new action, however the same thing happens quite soon after.
It also becomes impossible to refresh the page at this point. However closing a tab/window in the browser resets the problem.

**How to reproduce**  
Go to https://demo.api-platform.com/admin#/books and click the Author header about 7 times to sort the list by author.
You should see that the refresh button turns into the loading spinner. Now when you for example sort by a different column a few times it completely hangs and refreshing the page will no longer work.

alanpoulain:
It seems it's because each time a sorting is done, all SSE streams are created again without closing the previous ones and it saturates the browser.
Something needs probably to be changed in https://github.com/api-platform/admin/blob/main/src/ListGuesser.js.

4c0n:
It seems that subsciptions are managed in the data provider and a reference to the dataprovider seems available from
https://github.com/api-platform/admin/blob/main/src/useMercureSubscription.js

I'd say it would probably be a good idea to fix the problem from there. Maybe you or @Shywim have a better idea?
There's no documentation on how this is supposed to work, so it's a little hard for me to understand, I'm also by no means an expert on any of this.

However the whole Mercure feature just makes the admin very unusable at this point and the only way to get around it is to just disable it. Which is a shame, because it would be really nice to have.

PR #424, opened 2022-01-25
fix: Close event source connections

4c0n:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/admin/issues/413
| License       | MIT

As mentioned in the issue, the SSE streams for the Mercure feature are not closed before setting up new ones on rendering the page. That produces issues in the browser as there will be too many of them open (but not used) after a few user actions, leading to an unresponsive UI.

This change simply closes the streams.

### DIFF
--- a/src/hydra/dataProvider.ts
+++ b/src/hydra/dataProvider.ts
@@ -806,6 +806,7 @@ function dataProvider(
         if (sub.count <= 0) {
           if (sub.subscribed && sub.eventSource && sub.eventListener) {
             sub.eventSource.removeEventListener('message', sub.eventListener);
+            sub.eventSource.close();
           }
           delete subscriptions[resourceId];
         }


src/hydra/dataProvider.ts
```diff
@@ -806,6 +806,7 @@ function dataProvider(
         if (sub.count <= 0) {
           if (sub.subscribed && sub.eventSource && sub.eventListener) {
             sub.eventSource.removeEventListener('message', sub.eventListener);
```
alanpoulain, line 808: I'm wondering if this line is useful, WDYT?

4c0n: I'm guessing that the object should fall out of scope (no longer have any references) and will be removed after this (cleaned up by the garbage collector), which would make it not very useful. Maybe the original idea behind it was that the EventSource would be re-used by the browser instead of re-connecting and setting it up again, but I can't be sure.
I suppose it won't hurt to remove the event listener before closing though.

alanpoulain:
Thank you @4c0n.

PR #446, opened 2022-04-15
fix: change FieldGuesser key in ListGuesser when orderField is updated

alanpoulain:
This fix is needed to force the DatagridHeaderCell to render again when the order field is resolved and to enable sorting for the column.


src/ListGuesser.tsx
```diff
@@ -90,21 +92,23 @@ export const IntrospectedListGuesser = ({
...
           source={field.name}
           sortable={!!orderField}
           sortBy={orderField}
           resource={props.resource}
         />
       );
     });
-    displayOverrideCode(schema, readableFields);
+    if (orderParameters !== undefined) {
+      displayOverrideCode(schema, readableFields);
```
alanpoulain, line 110: To display the message only once.

PR #466, opened 2022-09-08
match https schema.org

### LINKED ISSUE #465: displaying name of relation in place of iri does not work for https://schema.org
**Description**  

When using `https://schema.org` displaying name in place of iri does not work, so this literal hint does not work (or any defaults generated from schema-generator):
https://api-platform.com/docs/admin/schema.org/#displaying-related-resources-name-instead-of-its-iri

only this one would:
```php
#[ApiProperty(iri: "http://schema.org/name")]
private $name;
```

Name hinting is hardcoded to `http://schema.org`:
https://github.com/api-platform/admin/blob/46b68715f1fd40e5916cd2e8d877808523b3b3b8/src/hydra/schemaAnalyzer.ts#L21

**Possible Solution**  
Match https also, i.e.
```js
  const field = schema.fields.find(
    (schemaField) => schemaField.id.match('https?://schema.org/name')
);
```

PawelSuwinski:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #465
| License       | MIT
| Doc PR        | No

Fixes #465


alanpoulain:
Could you fix the TS error please?

PawelSuwinski:
> Could you fix the TS error please?

Done.

alanpoulain:
Now it's Prettier :grin:

PawelSuwinski:
> Now it's Prettier grin
Heh, I omitted setting up whole development locally and running checks for such a small change, but it seems that quick fix is rarely quick :). Fixed.

PR #467, opened 2022-09-09
schemaAnalyzer.getFieldType scheme agnostic - match https and http

PawelSuwinski:
`schemaAnalyzer.getFieldType()` works only with `http://` scheme (similar case to #466). This makes it independent of the scheme.

### DIFF
--- a/src/hydra/schemaAnalyzer.ts
+++ b/src/hydra/schemaAnalyzer.ts
@@ -7,6 +7,9 @@ import {
 } from '../schemaAnalyzer';
 import type { SchemaAnalyzer, SubmissionErrors } from '../types';
 
+const withHttpScheme = (value: string | null | undefined) =>
+  value?.startsWith('https://') ? value.replace(/^https/, 'http') : value;
+
 /**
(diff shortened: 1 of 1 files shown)


src/hydra/schemaAnalyzer.ts
```diff
@@ -29,9 +29,13 @@ const getFieldNameFromSchema = (schema: Resource) => {
  * @returns The type of the field
  */
 const getFieldType = (field: Field) => {
-  switch (field.id) {
+  const withHttpScheme = (value: string | null | undefined) =>
```
alanpoulain, line 32: I think we should make the code uniform: either use this function for `getFieldNameFromSchema` too or use regular expressions here.

alanpoulain (approved):

src/hydra/schemaAnalyzer.ts
PawelSuwinski: Done

alanpoulain:
Thanks again @PawelSuwinski.

PR #474, opened 2022-09-16
fix reference input validation

PawelSuwinski:
Required referenced fields are not marked as required by InputGuesser:

For example field form parsed from hydra:
```js
{ 
  name: "sourceLanguage", 
  (...)
  embedded: null,
  required: true,
  maxCardinality: 1,
  deprecated: false
}
```
Looks like this:
obraz
SwaggerUI recognizes it correctly:
obraz

It seems that validation prop should be passed directly to Input field component, not to ReferenceInput wrapper. With that change all works fine.

### DIFF
--- a/src/InputGuesser.tsx
+++ b/src/InputGuesser.tsx
@@ -23,6 +23,8 @@ import type {
   NumberInputProps,
   ReferenceArrayInputProps,
   ReferenceInputProps,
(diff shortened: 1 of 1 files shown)


alanpoulain:
Thanks @PawelSuwinski.

PR #475, opened 2022-09-23
multiple file upload

PawelSuwinski:
Support of multiple files upload by dataProvider.

Example of usage:

```js
 <CreateGuesser {...props}>
    <FileInput source={"docs"} multiple={true} >
        <FileField source={"src"} title={"title"} download />
     </FileInput>
    (...)
 </CreateGuesser>
```
Some screenshot:
obraz

### DIFF
--- a/src/hydra/dataProvider.ts
+++ b/src/hydra/dataProvider.ts
@@ -255,11 +255,16 @@ function dataProvider(
       data as Record<string, unknown>,
     ).then((hydraData) => {
       const values = Object.values(hydraData);
-      const containFile = (element: unknown) =>
-        isPlainObject(element) &&
-        Object.values(element as Record<string, unknown>).some(
-          (value) => value instanceof File,
-        );
+      const containFile = (element: unknown): boolean =>
+        Array.isArray(element)
+          ? element
+              .map((value) => containFile(value))
+              .every((contain) => contain)
+          : isPlainObject(element) &&
+            Object.values(element as Record<string, unknown>).some(
+              (value) => value instanceof File,
+            );
+
       type ToJSONObject = { toJSON(): string };
       const hasToJSON = (
         element: string | ToJSONObject,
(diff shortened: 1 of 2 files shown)


src/hydra/dataProvider.ts
```diff
@@ -255,11 +255,16 @@ function dataProvider(
...
-      const containFile = (element: unknown) =>
-        isPlainObject(element) &&
-        Object.values(element as Record<string, unknown>).some(
-          (value) => value instanceof File,
-        );
+      const containFile = (element: unknown): boolean =>
+        Array.isArray(element)
+          ? element
+              .map((value) => containFile(value))
+              .every((contain) => contain)
```
alanpoulain, line 262: Can be simplified:
```suggestion
          ? element.every((value) => containFile(value))
```

src/hydra/dataProvider.ts
```diff
@@ -281,10 +286,17 @@ function dataProvider(
...
-            Object.values(value).find((val) => val instanceof File),
-          );
+          const findFile = (element: string | ToJSONObject): Blob =>
+            Object.values(element).find((val) => val instanceof File);
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          Array.isArray(value)
+            ? value
+                .map((val) => findFile(val))
+                .forEach((file) => {
+                  body.append(key.endsWith('[]') ? key : `${key}[]`, file);
```
alanpoulain, line 296: Is the ternary condition needed?
Could the key not end with `[]`?

alanpoulain:
Nice improvement!

src/hydra/dataProvider.ts
PawelSuwinski: Just to be meticulous, not harmful but can handle some corner cases, and the question is rather opposite - "could the key end with `[]`?".  Very seldom, only  when set by explicit.  To support nicely and transparently all resource labels, its translations,  etc.,  key should by passed without `[]` suffix to all InputGuesser/ InputField components, so it is done this way right now, and that why we need to transform it here adding this array `[]` suffix.

alanpoulain (approved):

alanpoulain:
Thank you @PawelSuwinski.

PR #476, opened 2022-09-26
feat: allow to use tabbed components in guessers

PawelSuwinski:
Extension that allows to use TabbedForm and TabbedShowLayout in place of SimpleForm and SimpleShowLayout.  Based on presence of tabbed element (FormTab / Tab)  appropriate type of form or show layout is choosen by guesser. 

https://marmelab.com/react-admin/doc/4.0/TabbedForm.html
https://marmelab.com/react-admin/doc/4.0/TabbedShowLayout.html

For example:
```
 <CreateGuesser {...props} >
      <FormTab label={"Tab 1"}>
        <InputGuesser source={"comments"} multiline  />
         (...)
      </FormTab>
     <FormTab label={"Tab 2"}>
         <InputGuesser source={"name"}
          (...)
      </FormTab>
</CreateGuesser>
```

Test case provided  for ShowGuesser for that moment.

### DIFF
--- a/src/CreateGuesser.tsx
+++ b/src/CreateGuesser.tsx
@@ -1,9 +1,12 @@
 import React, { useCallback } from 'react';
+import type { PropsWithChildren, ReactNode } from 'react';
 import PropTypes from 'prop-types';
 import {
   Create,
   FileInput,
+  FormTab,
   SimpleForm,
+  TabbedForm,
   useCreate,
   useNotify,
   useRedirect,
@@ -55,7 +58,7 @@ export const IntrospectedCreateGuesser = ({
   toolbar,
   warnWhenUnsavedChanges,
   sanitizeEmptyValues,
-  simpleFormComponent,
+  formComponent,
   children,
   ...props
 }: IntrospectedCreateGuesserProps) => {
@@ -77,10 +80,16 @@ export const IntrospectedCreateGuesser = ({
     displayOverrideCode(getOverrideCode(schema, writableFields));
   }
 
-  const hasFileField = inputChildren.some(
-    (child) =>
-      typeof child === 'object' && 'type' in child && child.type === FileInput,
-  );
+  const hasFileFieldElement = (elements: Array<ReactNode>): boolean =>
+    elements.some(
+      (child) =>
+        React.isValidElement(child) &&
+        (child.type === FileInput ||
+          hasFileFieldElement(
+            React.Children.toArray((child.props as PropsWithChildren).children),
+          )),
+    );
+  const hasFileField = hasFileFieldElement(inputChildren);
 
   const save = useCallback(
     async (values: Partial<RaRecord>) => {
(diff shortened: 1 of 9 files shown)


src/CreateGuesser.tsx
```diff
@@ -1,9 +1,12 @@
+import type { PropsWithChildren, ReactNode } from 'react';
```
alanpoulain, line 1: Could you add it after the "true" import?

src/CreateGuesser.tsx
```diff
@@ -77,10 +80,16 @@ export const IntrospectedCreateGuesser = ({
     displayOverrideCode(getOverrideCode(schema, writableFields));
   }
 
-  const hasFileField = inputChildren.some(
-    (child) =>
-      typeof child === 'object' && 'type' in child && child.type === FileInput,
-  );
+  const hasFileFieldElement = (elements: Array<ReactNode>): boolean =>
```
alanpoulain, line 83: Shouldn't this code be in the PR https://github.com/api-platform/admin/pull/475?

src/ShowGuesser.test.tsx
```diff
@@ -107,4 +112,34 @@ describe('<ShowGuesser />', () => {
...
+    );
+
+    await waitFor(() => {
+      expect(screen.queryAllByRole('tab')).toHaveLength(2);
+      // First tab.
+      expect(screen.queryAllByText('label of id')).toHaveLength(1);
+      expect(screen.queryAllByText('/users/123')).toHaveLength(1);
+      expect(screen.queryAllByText('label of title')).toHaveLength(1);
+      expect(screen.queryAllByText('Title')).toHaveLength(1);
+      // Second tab, unavailable.
```
alanpoulain, line 140: It would be nice to then click on the tab and verify that the body is now accessible.

src/types.ts
```diff
@@ -366,10 +367,12 @@ export type ResourceGuesserProps = Omit<
 >;
 
 type CreateSimpleFormProps = Omit<
-  CreateProps & Omit<SimpleFormProps, 'component'>,
+  CreateProps & SimpleFormProps & TabbedFormProps,
```
alanpoulain, line 370: Why the `component` prop is not omitted anymore?

alanpoulain:
Great!
If you find some motivation/time, it would be great to add integration tests for the `CreateGuesser` and `EditGuesser` components too.

src/CreateGuesser.tsx
PawelSuwinski: Nope, multiple file upload does not place  FileInput element deeper in the tree, it is still on the first level just with additional `multiple` attribute, so this works as it is:
```
  <CreateGuesser {...props} >
      <FileInput source={"files"} multiple={true} >
   (...)
```
But tabbed form place it deeper, so the there is a need to look deeper: 
```
  <CreateGuesser {...props} >
    <FormTab lable={"files"}>
      <FileInput source={"files"}  >
   (...)
```
And after all looking deeper it support any other nesting of FileInput component.

src/ShowGuesser.test.tsx
PawelSuwinski: OK, I will add it.

src/CreateGuesser.tsx
PawelSuwinski: OK

src/types.ts
PawelSuwinski: Is it not redundant here? 

Intersection with `Patial<PickRename>` to replace "component" with "formComponent" should do the thing. I just used as a pattern simplified syntax from `EditSimpleFormProps` to unify it both statements :
https://github.com/api-platform/admin/blob/3762ad0a8ffb300554df77b6804ce751e720c7e5/src/types.ts#L388

PawelSuwinski:
> If you find some motivation/time, it would be great to add integration tests for the `CreateGuesser` and `EditGuesser` components too.

Done, I added tests  for Create/EditGuesser regarding this improvement (tabbed form). Closely  based on ShowGuesser tests.

alanpoulain:
Thank you @PawelSuwinski!

PR #495, opened 2023-02-13
FieldGuesser: props passing to SingleFieldList  (linkType prop)

PawelSuwinski:
By default any reference field links to edit view. 

To change it RA uses  [`link` prop](https://marmelab.com/react-admin/ReferenceField.html#link) for single field and [`linkType` prop](https://marmelab.com/react-admin/SingleFieldList.html#linktype) for array.

So that works fine:
```
<FieldGuesser source={"user"} link={"show"} />
```

But for array it doesn't because nested `SimpeFieldList` component does't get its prop:
```
<FieldGuesser source={"users"} linkType={"show"} />
```
So the only solution for that is to skip using FieldGuesser for arrays and write it by hand.

IHMO it is good to unify that and support it in both cases by FieldGuesser especially that it is very common use case to have all relations linked to show pages.


src/FieldGuesser.tsx
```diff
@@ -69,7 +70,7 @@ const renderField = (
       <ReferenceArrayField
         {...(props as ReferenceArrayFieldProps)}
         reference={field.reference.name}>
-        <SingleFieldList>
+        <SingleFieldList {...(props as SingleFieldListProps)}>
```
alanpoulain, line 73: Maybe passing `linkType` explicitly would be better to avoid a props "mix".

alanpoulain:
CI seems to fail, I need some time to dig the issue, feel free to do it if you want :slightly_smiling_face:

src/FieldGuesser.tsx
PawelSuwinski: OK, changed.

PawelSuwinski:
Checks solved, so what about merging it?

dunglas:
Thank you!

PR #498, opened 2023-02-24
refactor: move extraInformation to meta

mkg20001:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #... 
| License       | MIT
| Doc PR        | 

The *Params datatypes have a meta field that can be filled with any value as necesarry for the dataprovider, see UpdateParams here https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/types.ts#L189-L194

It's better to use this value instead of polluting the data object


primal100:
This is a priority also because it is only removed before sending in the hydra dataProvider. With OpenAPI, extraInformation is sent with the request which is then rejected on the server side because of the additonal field.

PR #506, opened 2023-08-09
fixed: random error message about apiSchema being null

slim:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | 

You randomly get a message saying `apiSchema` is null. This happened when the window lost and retrieved focus *twice* because the first time it lost focus it reconnected to the api endpoint and re-instantiated `dataProvider` without updating the Schema by reconnecting to `/`. Next time it tries to connect to the api endpoint, `apiSchema` which was a local variable was already lost.

This bug was notably happening in api-platform demo

### DIFF
--- a/src/EditGuesser.tsx
+++ b/src/EditGuesser.tsx
@@ -197,7 +197,10 @@ export const IntrospectedEditGuesser = ({
       id={id}
       mutationMode={mutationMode}
       redirect={redirectTo}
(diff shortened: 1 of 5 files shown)


dunglas (approved):

dunglas:
Thanks for the fix! Could you add a non-regression test for that?

dunglas:
test failures look related

alanpoulain (approved):

alanpoulain:
Thank you Slim!

PR #510, opened 2023-10-05
compound order by

PawelSuwinski:
Hi, 

Adds possibility to sort of compound filed, ex.
```
const UserList = (props) => (
<ListGuesser {...props}>
    // ...
    <FunctionField
       label={"Full name"}
       sortBy={"lastName, firstName"}
        render={(record) => `${record.lastName} ${record.firstName}`}
    />
</ListGuesser>
)
```
On column sort it gives queries with multiple order parameters:
```
https://localhost/users?order[lastName]=asc&order[firstName]=asc
https://localhttps://localhost/users?order[lastName]=desc&order[firstName]=desc
```

API Platform Core and SwaggerUI supports  compound order (multiple order parameter), but Api Platform Admin doesn't allow to achieve it in some simple way.  AFAIK ReactAdmin itself doesn't plan to support multicolumn order in data grids components but resolving of single column order based on sortBy porp  it just a dataProvider job, and can be nicely extended to support such kind compound order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.4.5
+
+* Fix validation errors shown as "Server communication error" when creating an entity
+
 ## 3.4.4
 
 * Enum support in field guesser
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -1,8 +1,9 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import '@testing-library/jest-dom';
-import { TextEncoder } from 'util';
+import { TextEncoder, TextDecoder } from 'util';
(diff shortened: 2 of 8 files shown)


soyuka:
yes it supports that ? https://api-platform.com/docs/core/filters/#order-filter-sorting

PawelSuwinski:
> yes it supports that ? https://api-platform.com/docs/core/filters/#order-filter-sorting

Yes, backed does (Core) but it can't be used on the frontend side (excluding SwaggerUI) without extending API Platform Admin dataProvider. Correct me if I miss something.

PawelSuwinski:
Hi, what about merging it? I thinks it is quite useful, and covers gap between core and admin. We use it as a patch, but would be nice to have it official.

dunglas:
Thank you!

PR #514, opened 2023-11-07
feat: export themes

herndlm:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       |  
| License       | MIT
| Doc PR        | 

I like to use / override parts of the themes here and was wondering why it is not exported. Is there any reason not to do so?


dunglas (approved):

dunglas:
Thanks!

PR #515, opened 2023-11-08
Fixed #488 Validation errors are shown as "Server communication error" 

slim:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | 
| License       | MIT
| Doc PR        | 

This fixes a bug which is a combination of 2 bugs in [api-doc-parser library](https://github.com/api-platform/api-doc-parser) and 2 bugs in this repo. 

Fix also requires PR to be merged https://github.com/api-platform/api-doc-parser/pull/132

PR #528, opened 2024-02-19
null violations error body fix

PawelSuwinski:
I seems that PR #515 broke violations list processing based on error body so validation messages are not shown any more in the form (?), this would never happen:

https://github.com/api-platform/admin/blob/a3292d81ddab39c04152e239e641920dd6f156ff/src/hydra/fetchHydra.ts#L79-L81

I think that we should step back and let jsonld do the math like it was in previous version. With this fix validation errors are shown again in the form. Please correct me if I missed anything here.


dominikboniecki:
@alanpoulain please look at this change, the bug is very critical for global error handling in a whole project

dunglas:
Thank you!

PR #535, opened 2024-03-06
Docker storybook

soyuka:
Continues #518


src/hydra/dataProvider.ts
```diff
@@ -196,6 +196,7 @@ function dataProvider(
       }
     : null;
 
+  console.log('hi');
```
adguernier, line 199: console.log spotted

adguernier (changes requested):

PR #540, opened 2024-03-27
CreateGuesser / EditGuesser formId attribute to bind outher input elements

PawelSuwinski:
Possibility to set guesser's form id using `formId` attribute. It allows to place any [input element outside the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form). 

Example of submit button on the AppBar (just to show the idea, whole implementation uses  `React.createPortal()` and a bit more code):   

```
const AppBar  = () => (
   /...
  <SaveButton form="user-form" />
);
// ...

const UserCreate = () => (
    <CreateGuesser formId="user-form" (...)>
     // ...
    </CreateGuesser>
);
```
obraz

### DIFF
--- a/src/CreateGuesser.tsx
+++ b/src/CreateGuesser.tsx
@@ -58,6 +58,7 @@ export const IntrospectedCreateGuesser = ({
   toolbar,
   warnWhenUnsavedChanges,
   sanitizeEmptyValues = true,
+  formId,
   formComponent,
   viewComponent,
   children,
@@ -165,6 +166,7 @@ export const IntrospectedCreateGuesser = ({
   return (
     <Create resource={resource} component={viewComponent} {...props}>
       <FormType
+        id={formId}
         onSubmit={save}
         mode={mode}
         defaultValues={defaultValues}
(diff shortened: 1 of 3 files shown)


fzaninotto:
In my opinion, the Guesser should only get you started, and then you should paste its code and customize it. Otherwise, it will become some kind of "god component", which goes against the philosophy of react-admin.

PawelSuwinski:
> In my opinion, the Guesser should only get you started, and then you should paste its code and customize it. Otherwise, it will become some kind of "god component", which goes against the philosophy of react-admin.

In generall you are right but api-platform Guessers are something more than RA Guessers. There is a override-code message that hints the customization (https://github.com/api-platform/admin/blob/4645fb57ad8a9b443321677ba3cd3fe9c2e8cb33/src/CreateGuesser.tsx#L26) but there are also introsopection, transform and other purpouse callbacks that form needed, and it was the goal of the api-platform, not to assembly it by hand. Maybe it should be named in other way then guessers,  or all these helper callbacks should be exported as hooks, but now it is not.

PawelSuwinski:
@fzaninotto your comment made me want to do some refactoring (#542) to makes easier to write custom form components, so please give some feedback about that.

PR #542, opened 2024-03-28
useOnSubmit hook for Create/ EditGuesser

PawelSuwinski:
Refactored form submit handler and moved to stand alone hook. It makes easier to write custom form components with introspection without copy/ paste of `save` callback code from Guessers components. 

See #540.

### DIFF
--- a/src/CreateGuesser.tsx
+++ b/src/CreateGuesser.tsx
@@ -1,23 +1,18 @@
-import React, { useCallback } from 'react';
-import type { PropsWithChildren, ReactNode } from 'react';
+import React from 'react';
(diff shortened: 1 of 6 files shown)


fzaninotto:
This does indeed look like a progress in comparison with the previous version. 

However, I still think that you should package the logic into a component that developers can paste and then modify. If you follow the react-admin conventions, the Guesser should just just render a series of components and paste them to the console so that the developer can customize them. It seems that part of the logic you've put in the guessers should be in some specific HydraEdit, HydraCreate and HydraList components.

But I understand it's not the purpose of this PR, so LGTM.

PR #554, opened 2024-05-14
fix: pwa container crashes because packageManager is not defined

adguernier:
## Problem

Running `docker compose up` throw the error: "The local project doesn't define a 'packageManager' field" for the pwa container

## Solution

Specify a `packageManager` in `package.json`


alanpoulain (approved):

alanpoulain:
It seems like the CI is not happy with it.

slax57 (approved):

Issue #551, opened 2024-05-02
Custom AuthProvider not working with <HydraAdmin/>


tmihalicka:
**API Platform version(s) affected**: 3.4.6

**Description**  
After upgrading `@api-platform/admin` package to version `3.4.6` from version `3.4.5`, authProvider not working when is used with `<HydraAdmin />`, when i downgrade package to version `3.4.5` everything start works flawlessly

**How to reproduce**  
Add custom `authProvider` to `<HydraAdmin />`

**Additional Context**  
My `<HydraAdmin />` init
```javascript
const App = () => {
  return (
    <Sentry.ErrorBoundary fallback={'An error has occurred'}>
      <HydraAdmin
        entrypoint={APP_CONSTANTS.ENTRYPOINT}
        dataProvider={dataProvider}
        authProvider={authProvider}
        customRoutes={customRoutes}
        theme={lightTheme}
        lightTheme={lightTheme}
        darkTheme={darkTheme}
        layout={AdminLayout}
        disableTelemetry={true}
        title={'Admin'}
      >
        {
          permissions => typeof permissions !== 'undefined' ? resourceComponents.map((component) => {
            const { resources } = permissions;
            return resources.includes(component.key) ? component : null;
          }) : <React.Fragment />
        }
      </HydraAdmin>
    </Sentry.ErrorBoundary>
  );
};
```
auth provider is not called, so I don't have information about the logged user and permission for this user.

7system7:
I have the same issue.

papppeter:
Same issue for me.

7system7:
It seems, the problem is much older... I have a ~6 years old project, and when I delete everything and rebuild the whole FE

```bash
$ pnpm store prune && rm -rf node_modules pnpm-lock.yaml && pnpm i && pnpm dev
```

the auth-provider has never been called anymore. Not w/ the latest, nor w/ the 3.0.0. :disappointed:

If I 
 - restore the original (before this mods) `pnpm-lock.yaml` and reinstall the packages, **the auth-provider works**.
 - I use the `react-admin`'s Admin component, **the auth-provider works**.

PawelSuwinski:
https://github.com/api-platform/admin/issues/550

7system7:
@PawelSuwinski Thank u! It is working.

7system7:
Sorry, I was on holiday, and I just quickly checked this solution.

[...]

PR #558, opened 2024-05-22
AdminGuesserProps - missing admin prop type

PawelSuwinski:
Overriding admin element by  `admin` prop in HydraAdmin/ OpenApiAdmin  (for example like here: https://github.com/api-platform/admin/issues/551#issuecomment-2095514796) works as expected however tsc compilation throws  an error: `Property 'admin' does not exist on type (...)`. This is the fix.


src/AdminGuesser.test.tsx
```diff
@@ -83,4 +86,36 @@ describe('<AdminGuesser />', () => {
...
+
+    renderer.render(
+      <AdminGuesser
+        admin={AdminEl}
+        dataProvider={dataProvider}
+        schemaAnalyzer={schemaAnalyzer()}
+      />,
+    );
+
+    expect(renderer.getRenderOutput()).toMatchSnapshot();
```
alanpoulain, line 119: Maybe matching the snapshot is too much.

PawelSuwinski: Indeed, I'll change it.

alanpoulain (approved):

Issue #561, opened 2024-05-29
InputGuesser - incomplete typing for reference input


PawelSuwinski:
**API Platform Admin version(s) affected**: 3.4.7

InputGuesser for reference field with some extra props of ReferenceInput(Array) like sort or filter ends with typescript error, ex.:
```
<InputGuesser
  source={"owner"}
  sort={{ field: "id", order: "DESC" } as SortPayload}
/>
```
ends with:
```
(...)
    Property 'sort' does not exist on type 'IntrinsicAttributes & Pick<InputGuesserProps, "id" | "title" | "type" | "required" | "name" | "key" | "ref" | "label" | "slot" | "source" | ... 279 more ... | "transformEnum"> & Pick<...> & Pick<...>'.

    256  <InputGuesser source="owner" sort={{field: "id", order: "DESC" }} />
```

It looks like that InputGuesserProps type is somehow incomplete. I have some draft PR with test case (see linked), but can't get it to work (some RA issue?).

PR #562, opened 2024-05-29
InputGuesser - incomplete typing for reference input

PawelSuwinski:
Adds missing types to InputGuesserProps to allow usage of ReferenceInput/ ReferenceInputArray extra props (ex. `sort` or `filter`).

However test case still fails with _(updated, see below)_:
```
Property 'sort' does not exist on type (...)
```
Am I missing something (?)... I will come back to it when I have more time to dig it.

Fix: #561

### DIFF
--- a/src/InputGuesser.test.tsx
+++ b/src/InputGuesser.test.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
+import type { SortPayload } from 'react-admin';
 import {
   AdminContext,
   Edit,
   ResourceContextProvider,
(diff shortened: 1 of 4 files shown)


adguernier (approved):

fzaninotto:
Merging this PR broke the `main` branch (see https://github.com/api-platform/admin/actions/runs/9647105787/job/26605107106). Can you revert or fix the build?

soyuka:
this was green before merging

PR #575, opened 2024-07-27
Fix default theme and store

fzaninotto:
## Problem

The default HydraAdmin differs from react-admin in 2 ways:

1. The default store is memoryStore instead of localStorageStore. As a consequence, filters and sort preferences aren't persisted between sessions (and aren't auditable in devtools)
2. The inputs aren't legible in dark theme because they lack a background color

Also, it is not possible to customize the darkTheme.

## Solution

Use similar defaults as react-admin

| Before | After |
| --- | --- |
| image|image|

| Q             | A
| ------------- | ---
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+* Fix Inputs background in darkTheme
+* Fix default store isn't using localStorage
+
 ## 4.0.0
 
 * Compatibility with react-admin v5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 # Changelog
 
-* Fix Inputs background in darkTheme
-* Fix default store isn't using localStorage
-
 ## 4.0.0
 
 * Compatibility with react-admin v5
--- a/src/core/AdminGuesser.tsx
+++ b/src/core/AdminGuesser.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useMemo, useState } from 'react';
 import {
   AdminContext,
   defaultI18nProvider,
-  localStorageStore,
+  /* tree-shaking no-side-effects-when-called */ localStorageStore,
 } from 'react-admin';
 
 import type { ComponentType } from 'react';
--- a/src/core/AdminGuesser.tsx
+++ b/src/core/AdminGuesser.tsx
@@ -1,5 +1,9 @@
 import React, { useEffect, useMemo, useState } from 'react';
-import { AdminContext, defaultI18nProvider } from 'react-admin';
+import {
+  AdminContext,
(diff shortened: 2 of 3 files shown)


src/layout/themes.ts
```diff
@@ -34,6 +34,9 @@ export const darkTheme: RaThemeOptions = {
         },
       },
     },
+    MuiFilledInput: {
```
fzaninotto, line 37: The custom `darkTheme` here extends react-admin's `defaultTheme`, which is in fact the `defaultLightTheme` and contains light-theme specific changes for inputs that should not apply to dark theme.

src/core/AdminGuesser.tsx
```diff
@@ -98,7 +105,6 @@ const AdminGuesser = ({
       queryClient={queryClient}
       theme={theme}
       darkTheme={darkTheme}
-      lightTheme={lightTheme}
```
fzaninotto, line 101: `lightTheme` isn't documented in react-admin - `theme` is the prop to use. 

https://marmelab.com/react-admin/Admin.html#theme

src/core/AdminGuesser.tsx
```diff
@@ -42,7 +48,8 @@ const AdminGuesser = ({
   layout = Layout,
   loginPage = LoginPage,
   loading: loadingPage,
-  theme = lightTheme,
+  theme = defaultLightTheme,
+  darkTheme = defaultDarkTheme,
```
fzaninotto, line 52: There was a bug here - the dark theme couldn't be overridden

src/core/AdminGuesser.tsx
```diff
@@ -33,7 +39,7 @@ const AdminGuesser = ({
   basename,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   error = DefaultError as any,
-  store,
+  store = defaultStore,
```
fzaninotto, line 42: React-admin defines the default store in `react-adminù 's `Admin` component. Since `AdminGuesser` relies on the lower-level `AdminContext`, we have to repeat the same initialization.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+* Fix Inputs background in darkTheme
```
adguernier, line 3: There is no need to update the CHANGELOG as it will be automatically updated upon the release

adguernier (changes requested):
Code looks good to me :+1:

CHANGELOG.md
fzaninotto: Fixed

adguernier (approved):

PR #585, opened 2024-10-07
fix : hydra

valentin-dassonville:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

Up to API Platform 4.

Get the error : ```Cannot find the class related to https://localhost/docs.jsonld#Entrypoint/greeting``` from api-doc-parser.

@dunglas @soyuka

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ If you already have a project in progress, you can develop directly from it.
 
 The instructions below explain how to install the source version of API Platform Admin in your project and contribute a patch.
 
-Your client should already use `@api-platform/admin` and its bootstrap file (usually: `src/App.tsx`) should at least contains: 
+Your client should already use `@api-platform/admin` and its bootstrap file (usually: `src/App.tsx`) should at least contains:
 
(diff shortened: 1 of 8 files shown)


valentin-dassonville:
The problem seems to be from api-doc-parser.
The key for `return` is wrong.
This is `supportedOperation` in a test that i made.
```
"http://www.w3.org/ns/hydra/core#supportedOperation": [
      {
        "@type": [
          "http://www.w3.org/ns/hydra/core#Operation",
          "https://schema.org/FindAction"
        ],
        "http://www.w3.org/2000/01/rdf-schema#label": [
          {
            "@value": "Retrieves the collection of Greeting resources."
          }
        ],
        "http://www.w3.org/ns/hydra/core#method": [
          {
            "@value": "GET"
          }
        ],
        "http://www.w3.org/ns/hydra/core#title": [
          {
            "@value": "Retrieves the collection of Greeting resources."
          }
        ],
        "https://localhost/docs.jsonld#returns": [
          {
            "@id": "https://localhost/Collection"
          }
        ]
      }
```

Issue #370, opened 2021-02-26
Using PATCH for update / edit


Antarian:
**Description**  
In this line in code
https://github.com/api-platform/admin/blob/5badc1f19024c1800a60431e575219039b86d0b6/src/hydra/dataProvider.js#L340
is forced PUT for update operation. s it posible to somehow detect if schema is using `put` or `patch`? JsonLD is displaying this information.

alanpoulain:
React admin always sends all the data, this behavior would need to be changed beforehand for this feature to be useful.

Antarian:
yes, but specific PATCH format (merge-patch+json) may send all the data. And I think I found something about this partial PATCH in React Admin issues. Will send link here in a moment.
Maybe some kind of manual config here would be useful, but that is sounding kind of like a hack. As ApiPlatform is creating new child entity when PUT is used and overriding same entity when PATCH is used. Which is my required and expected behaviour for EDIT.
So far I retreat to hack/decorate serializer/normalizer to alter this behaviour of PUT in ApiPlatform when `id` is sent. If I will be successful.

Antarian:
https://github.com/marmelab/react-admin/issues/256
and also this one, but not sure how much is it related:
https://github.com/josx/ra-data-feathers/issues/20
with this PR
https://github.com/josx/ra-data-feathers/pull/49/files

vaheeds:
I am using react-admin as the front-end. In my case, the backend API uses some sort of validation and does not allow `_id` to be modified. So it cannot accept PUT requests and throws an error when the `_id` field is presented in the PATCH request. So I have to use PATCH and I want to send modified fields only.
How can I overcome this issue?

vaheeds:
I figured it out in a custom data provider:

```
  update: (resource, params) =>
    httpClient(`${apiUrl}/${resource}/${params.id}`, {
      method: 'PATCH',
      body: JSON.stringify(diff(params.data, params.previousData)),
    }).then(({ json }) => ({ data: json })),
```

where the `diff` function is:

```
const diff = (object, base) => {
  return transform(object, (result, value, key) => {
    if (!isEqual(value, base[key])) {
      result[key] =
        isObject(value) && isObject(base[key]) ? diff(value, base[key]) : value;
    }
  });
};
```
now, only the altered fielt sent in request body with a `PATCH` method.

vincentchalamon:
[...]

PR #589, opened 2024-11-08
fix: use patch by default

soyuka:
API Platform 4 now uses PATCH by default, PUT is not enabled anymore and needs to be enabled by-hand on the API. This PR changes the default value for our Hydra data provider.


vincentchalamon:
Fixes https://github.com/api-platform/admin/issues/370

dunglas (approved):
Good idea. That could be nice to make this configurable, but this can be done later in another PR.

soyuka:
Storybook tests are fine locally, still bugged on the CI I need to inestigate at some point.

PR #597, opened 2025-03-03
Fix local environment and CI job following upgrade to `api-platform/core` v4

slax57:
## Problems

- The `composer.lock` file is not up to date with the `composer.json` file, hence the local env is still effectively using `api-platform/core` v3 (see https://github.com/api-platform/admin/commit/c0879d2ee233477a37710dcbb7628950ff4bad03 and https://github.com/api-platform/admin/commit/6a1f26a7b6baf21a4ee4f881d9039d7d9f79a51e)
- Starting Storybook locally fails because of the attempt to open a browser using `xdg-open`.
- The CI job **Storybook** currently fails on `main`: https://github.com/api-platform/admin/actions/runs/13050713440/job/36410149884

## Solution

- Update the `composer.lock`
- Update the Storybook container to start Storybook using the `--no-open` option
- Update `docker/bake-action` to v5

### DIFF
--- a/.github/workflows/storybook.yml
+++ b/.github/workflows/storybook.yml
@@ -24,7 +24,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
       -
           name: Build Docker images
-          uses: docker/bake-action@v4
+          uses: docker/bake-action@v5
           with:
               pull: true
               load: true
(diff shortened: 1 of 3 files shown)


soyuka:
Thanks!

PR #599, opened 2025-03-06
Fix infinite Loading progress on React 19

### LINKED ISSUE #598: [Bug] The project has never-ending Loading progress on React 19 (?)
**API Platform version(s) affected**: 4.0.17

**Description**  

We have a complex but well maintained project since 2019. We never have any problem with the major upgrades w/ MUI, react-admin nor w/ the React. But the upgrade to the React 19 and MUI v6 seems impossible. After I solve every errors and breaking changes on console and in the code I got this all the time:

Image

What is happening when this loading page is occurring? What the program waiting for exactly? It seems, the API is reachable and it gives all required information.

Image

or

Image

After a while, I realized that, if I install this project from scratch it fails the same way. I installed as the [docs said](https://api-platform.com/docs/admin/getting-started/). The code can be downloaded below.

[my-admin.zip](https://github.com/user-attachments/files/19094581/my-admin.zip)

I do not know what causing the problem; the react 19 or the MUI v6 or something else.

**How to reproduce**  

1. Install the app by the docs info. 
2. Run the project
3. See the never-ending circular loader...

SebLevDev:
Closes issue #598 - Fix infinite Loading progress on React 19

| Q             | A
| ------------- | ---
| Tickets       | Closes #598
| License       | MIT

### DIFF
--- a/src/core/AdminGuesser.tsx
+++ b/src/core/AdminGuesser.tsx
@@ -99,7 +99,7 @@ const AdminGuesser = ({
   if (loading) {
     return <Loading />;
   }
-  
+
   return (
     <AdminContext
       i18nProvider={i18nProvider}
--- a/src/core/AdminGuesser.tsx
+++ b/src/core/AdminGuesser.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import {
   AdminContext,
+  Loading,
   defaultI18nProvider,
   /* tree-shaking no-side-effects-when-called */ localStorageStore,
 } from 'react-admin';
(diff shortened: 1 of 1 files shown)


src/core/AdminGuesser.tsx
```diff
@@ -95,6 +96,10 @@ const AdminGuesser = ({
     [setLoading, setIntrospect],
   );
 
+  if (loading) {
+    return React.createElement(Loading, null);
```
dunglas, line 100: Maybe could we use JSX?

```suggestion
    return <Loading />;
```

soyuka:
thanks @SebLevDev

SebLevDev:
s it possible to publish a release?

PR #617, opened 2025-04-28
fix broken demo link

tacman:
link was broken


slax57:
Seems already applied. Closing.

PR #629, opened 2025-06-08
fix: Also apply sort params to existing fields in a ListGuesser

### LINKED ISSUE #626: Various API Filters not showing in the UI
**API Platform version(s) affected**: 4.1.0

**Description**  
Many of the available Filters seem to not be properly supported by the PWA.

**How to reproduce**  

I have an issue with the HydraAdmin component of the API-Platform distribution:

The following SearchFilter and OrderFilter attributes with nested properties are configured for the Coffee entity:

[![enter image description here][1]][1]

**The backend part works as expected**, so I can order and search by the given nested properties on the API itself. Also when I manually edit the respective query params in the HydraAdmin URL, it properly shows the searched and ordered results.

Here using `beans.type=/bean_types/{id}`
[![enter image description here][2]][2]

And here using `beans.type.name={name}`
[![enter image description here][3]][3]

The actual problem seems to be in the HydraAdmin so that **it does not offer the nested properties in the "Add Filter" section and also it does not show it under "Active Filters"**.

A similar but a bit different problem happens for the OrderFilter with nested properties: When I manually set it in the query params it works, but **when clicking on the roaster-order it sends `order=roaster` instead of `order=roaster.name`**

[![enter image description here][4]][4]

**Another broken case is the RangeFilter.** 

[![enter image description here][5]][5]

It passes the filter properly to the API but **does not show it under add filters or active filters**.

[![enter image description here][6]][6]

**Am I doing something wrong? Is there some (undocumented) way to properly integrate these filters in the UI? Or is it bugged or not yet supported by the UI?**

  [1]: https://i.sstatic.net/EwAvldZP.png
  [2]: https://i.sstatic.net/y4aYEm0w.png
  [3]: https://i.sstatic.net/FyHSq31V.png
  [4]: https://i.sstatic.net/K4kkaNGy.png
  [5]: https://i.sstatic.net/OlOkeYr1.png
  [6]: https://i.sstatic.net/4h2GuQ4L.png

zolex:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | partially fixes #626
| License       | MIT
| Doc PR        | -

Also apply sort params to existing fields in a ListGuesser instead of only "empty" listguessers.

### DIFF
--- a/src/list/ListGuesser.tsx
+++ b/src/list/ListGuesser.tsx
@@ -1,10 +1,30 @@
+import type { ReactElement } from 'react';
 import React, { useEffect, useState } from 'react';
 import {
+  ArrayField,
+  BooleanField,
+  ChipField,
   Datagrid,
   DatagridBody,
+  DateField,
   EditButton,
+  EmailField,
+  FileField,
+  FunctionField,
+  ImageField,
   List,
+  NumberField,
+  ReferenceArrayField,
+  ReferenceField,
+  ReferenceManyField,
+  ReferenceOneField,
+  RichTextField,
+  SelectField,
   ShowButton,
+  TextField,
+  TranslatableFields,
+  UrlField,
+  WrapperField,
   useResourceContext,
   useResourceDefinition,
 } from 'react-admin';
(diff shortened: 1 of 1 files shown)


slax57 (changes requested):
Thanks for offering to contribute in any case!

PR #630, opened 2025-06-08
fix: Don't show time in guessed date fields.

zolex:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -

Don't show time in guessed date fields.

### DIFF
--- a/src/field/FieldGuesser.tsx
+++ b/src/field/FieldGuesser.tsx
@@ -121,6 +121,8 @@ const renderField = (
       return <BooleanField {...(props as BooleanFieldProps)} />;
 
     case 'date':
+      return <DateField showTime={false} {...(props as DateFieldProps)} />;
+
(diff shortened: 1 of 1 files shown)


zolex:
> Thank you for the PR.
> 
> However `false` is already the default value for `showTime`.
> 
> Why did you need that change? Can you provide a way to reproduce the issue you are facing?

well, just use apip-distribution and create a model with a date field. it will show a datetime input.

slax57:
> well, just use apip-distribution and create a model with a date field. it will show a datetime input.

I did not reproduce using the example project included in this repo (which you can run by cloning and running `docker compose up`).
Do you reproduce the issue in there? :thinking: 
If I need to change the entities to reproduce can you tell me what change is needed?
Thanks

PR #658, opened 2026-03-26
Fix Create and Edit forms shows "Server communication error" instead of the root error message in case of validation error

### LINKED ISSUE #657: useOnSubmit shows "Server communication error" notification even when submission errors (including root.serverError) are returned
**API Platform version(s) affected**

- `@api-platform/admin`: 4.0.9
- `react-admin`: 5.14.x

**Description**

When a server returns submission errors (including `root.serverError`), the default `failure` handler in `useOnSubmit` always shows a generic `ra.notification.http_error` ("Server communication error") notification — even though `submissionErrors` are successfully extracted and returned to the form.

React Admin's `useNotifyIsFormInvalid` hook is designed to display the `root.serverError` message from form state, but the generic notification from `useOnSubmit` fires first, resulting in a confusing user experience.

This is related to #488.

**How to reproduce**

1. Configure `getSubmissionErrors` in the schema analyzer to return errors from the server response (required for OpenAPI setups, already built-in for Hydra)
2. Return a response with `root.serverError`:
   ```json
   {
     "errors": {
       "root": { "serverError": "Custom error message" },
       "title": "Title is required"
     }
   }
   ```
3. Submit the form
4. **Expected:** Notification shows "Custom error message" (from `useNotifyIsFormInvalid`)
5. **Actual:** Notification shows "Server communication error" (from the default `failure` handler in `useOnSubmit`)

**Root cause**

In `useOnSubmit.ts`, the default failure handler (lines 94–110):

```typescript
const failure =
  mutationOptions?.onError ??
  ((error: string | Error) => {
    let message = 'ra.notification.http_error';
    if (!submissionErrors) {
      message =
        typeof error === 'string' ? error : error.message || message;
    }
    // ...
    notify(message, {
      type: 'warning',
      messageArgs: { _: errorMessage },
    });
  });
```

When `submissionErrors` is truthy, the `if (!submissionErrors)` block is skipped, so `message` stays as `'ra.notification.http_error'`. The `notify()` call still executes, showing the generic error.

**Possible Solution**

Add an early return when `submissionErrors` exists, so the notification is left to React Admin's `useNotifyIsFormInvalid`:

```typescript
const failure =
  mutationOptions?.onError ??
  ((error: string | Error) => {
    if (submissionErrors) {
      return;
    }
    let message = 'ra.notification.http_error';
    message =
      typeof error === 'string' ? error : error.message || message;
    // ...
    notify(message, {
      type: 'warning',
      messageArgs: { _: errorMessage },
    });
  });
```

PawelSuwinski:
Let RA handle notification of validation errors to not show misleading "Server communication error" message.

#Fix #657


src/useOnSubmit.ts
```diff
@@ -92,11 +92,14 @@ const useOnSubmit = ({
...
-              message =
-                typeof error === 'string' ? error : error.message || message;
+            // Notification will be handled by the useNotifyIsFormInvalid hook.
+            if (submissionErrors) {
+              return;
             }
+            const message =
+              typeof error === 'string'
+                ? error
+                : error.message || 'ra.notification.http_error';
```
slax57, line 103: Thank you for the PR!

You may want to double check what happens in mutation modes other than pessimistic. Indeed, in such cases, react-admin might redirect the user optimistically out of the form, so the `useNotifyIsFormInvalid` hook will not trigger as it gets unmounted.

You can compare your implementation to the one in RA:

https://github.com/marmelab/react-admin/blob/2323d0b3ad2ead10c1edc71195ff3a6915c26cf3/packages/ra-core/src/controller/edit/useEditController.ts#L202-L212

PawelSuwinski: You are right, thanks for catching this. So I've fixed by following RA.

slax57 (approved):

